PR #1, opened 2020-04-14
pandas.DataFrame.ix is deprecated.

NTedjosantoso:
An AttributError 'DataFrame' object has no attribute 'ix' comes up if a real turbine power curve from windpower.Turbinelibrary is used for the simulation. (as in example 3.05d)

As of pandas 1.0 it is no longer possible to use pandas.DataFrame.ix.
Changing .ix to .loc within the code seems to be enough to fix the problem.


dsryberg:
Hey @NTedjosantoso, thanks for finding this!

PR #21, opened 2020-05-31
Update _powerCurveConvoluter.py

nicholastedjosantoso:
Hi. I found that a small change to the code is necessary. Otherwise it is not possible with the current version of pandas (Version 1.0 in my case) to convolute a real power curve from the turbine database. 

Removing np.array (row 11 of this code) seems to be enough to fix the problem.

Minimal example code to reproduce the problem:

from reskit import windpower
windpower.convolutePowerCurveByGuassian('N117_Delta')

Hope this can help :)


nicholastedjosantoso:
Sorry i used wrong account. I will resubmit the pull request

PR #23, opened 2020-06-08
DOCStrings

nlandwehrfzj:


sevberg:
Hey @nlandwehrfzj , 

I'm not sure this is right since none of these commits are dealing with DocStrings. Can you please double-check?

nlandwehrfzj:
I think the problem was that I did not send a pull request for my restruct branch, but for the master branch. I have send you a new pull request with my restruct branch. Could you check if you received that pull request? :)

PR #167, opened 2025-07-11
92 simplify and update offshore cost model

mstargardt:
Added function to scale offshore wind power plants CapEx depending on wind speed, distance to the coastline and water depth


mstargardt:
wasnt ready yet

PR #169, opened 2025-07-11
92 simplify and update offshore cost model

### LINKED ISSUE #156: simplify and update offshore cost model
This issue was originally created by: @chrisjwin

mstargardt:
Added function to scale given OffshoreCapex regarding, wind speed, water depth and distance to the nearest coastline. 
Test are included


reskit/default_paths.yaml
```diff
@@ -1,3 +1,6 @@
 turbine_library_path: Null
 baseline_onshore_turbine_definition_path: Null
-baseline_offshore_turbine_definition_path: Null
\ No newline at end of file
+baseline_offshore_turbine_definition_path: Null
+waterdepthFile: "/benchtop/shared_data/General_Bathymetric_Chart_of_the_Oceans_GEBCO/GEBCO_2020/GEBCO_tiles/"
+distancetoCoast: "/benchtop/projects/2021-m-stargardt-phd/02_GHR_2025/01_offshoreTiffs/GMT_intermediate_coast_distance_01d.tif"
+
```
phil-fzj, line 6: no hardcoded paths in public versions

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+    Args:
+        distance (float): Distance to coastline in kilometers.
+        capacity (float): Power plant's capacity in kW.
+        variableCostFactor (float, optional): Cost multiplier in €/MW/km (default = 1.35).
+        fixedCost (float, optional): Fixed connection cost (default = 0).
+
+    Returns:
+        float: Total cable connection cost in monetary units.
+
```
phil-fzj: "monetary cost units" is quite unspecific. the args however suggest that the unit would be €/MW/km?

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    mainframeMass = 1.228 * np.power(rd, 1.953)
+    mainframeCost = 627.28 * np.power(rd, 0.85)
+    platformAndRailingMass = 0.125 * mainframeMass
+    platformAndRailingCost = platformAndRailingMass * 8.7
+
+    electricalConnectionCost = cp * 40
+    hydraulicAndCoolingSystemCost = cp * 12
+    nacelleCost = 11.537 * cp + 3849.7
+    towerMass = 0.2694 * sa * hh + 1779
+    towerCost = towerMass * 1.5
```
phil-fzj: having units in the docstring is nice, but i think it is a lot easier to follow if one would have the respective units of each calculation/variable as a comment next to them

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        longitude (float): Longitude in decimal degrees.
+        waterDepthFolderPath (str, optional): Path to the folder containing water depth .tif files.
+
+    Returns:
+        float: Water depth at the specified location in meters (positive value). Returns None if not found.
+    """
+    depthFiles = glob.glob(os.path.join(waterDepthFolderPath, "*.tif"))
+    resultDepth = getRasterValueFromTifs(depthFiles, latitude, longitude)
+
+    return abs(resultDepth) if resultDepth is not None else None
```
phil-fzj: no used anywhere in reskit

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+    try:
+        value = gk.raster.interpolateValues(path, (longitude, latitude))
+
+        return value
+
+    except Exception as e:
+        print(f"Error at Lat: {latitude}, Lon: {longitude}: {e}")
+    return None
+
```
phil-fzj: not used anywhere in reskit

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        litValueAvgDistCoast, baseCap
+    )
+    newCableCost = cableCostBase * cableRatio
+
+    # Combine all costs
+    totalOffshoreCapex = (
+        newTurbineCost + newFoundationCost + newCableCost + overheadCostBase
+    )
+
+    return totalOffshoreCapex
```
phil-fzj: also this function would profit from having the respective unit next to the variables

phil-fzj (changes requested):
I added some comments @mstargardt 
Also we need the pipeline running for RESKit on Github before we can merge anything @julian-belina

reskit/default_paths.yaml
mstargardt: i will delete it

reskit/wind/economic/scale_Offshore_Capex.py
mstargardt: Yes

reskit/wind/economic/scale_Offshore_Capex.py
mstargardt: I will have to look it up in the cited sources,. Besides minor change, this cost function is directly copied from onshore_cost_model.py -> provided by earlier RESKit version

reskit/wind/economic/scale_Offshore_Capex.py
mstargardt: Delivers required input parameter for cost function

reskit/wind/economic/scale_Offshore_Capex.py
mstargardt: Delivers required input parameter for cost function

reskit/default_paths.py
```diff
@@ -4,3 +4,4 @@
 DEFAULT_PATHS = {}
 with open(join(dirname(__file__), "default_paths.yaml")) as f:
     DEFAULT_PATHS = yaml.safe_load(f)
+pass
```
chrisjwin: unnecessary

reskit/default_paths.yaml
```diff
@@ -1,3 +1,6 @@
 turbine_library_path: Null
 baseline_onshore_turbine_definition_path: Null
-baseline_offshore_turbine_definition_path: Null
\ No newline at end of file
+baseline_offshore_turbine_definition_path: Null
+waterdepthFile: "/benchtop/shared_data/General_Bathymetric_Chart_of_the_Oceans_GEBCO/GEBCO_2020/GEBCO_tiles/"
+distancetoCoast: "/benchtop/projects/2021-m-stargardt-phd/02_GHR_2025/01_offshoreTiffs/GMT_intermediate_coast_distance_01d.tif"
```
chrisjwin: Do we plan to have a default water depth and distance-to-coast file with the RESkit repo? Also @phil-fzj.  In that case, it should be added to the data folder (@julian-belina has implemented a nice way to deal with large data in geokit, will probably be applied here as well?) or by default this should be null and throw an error in the function then, saying that the file locations shall be specified in the default_paths.yaml. What we should not do is have our cluster paths in the open repo, since people without cluster access will definitely run into an error here

chrisjwin: I missed the comment below, nevertheless the discussion may be needed fyi @phil-fzj @mstargardt

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+def waterDepthFromLocation(
+    latitude,
+    longitude,
+    waterDepthFolderPath=DEFAULT_PATHS.get("waterdepthFile"),
+):
+    """
+    Returns the water depth (in meters) at a given geographic location (latitude and longitude).
+
+    Args:
```
chrisjwin: please adapt docstring to the style used elsewhere in RK

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+# %% function to calculate the distance to the coastline
+# if you want to execute the distance to coastline more often, please separete the loading of the taserband to increase execution time
+
+
+def distanceToCoastline(latitude, longitude, path=DEFAULT_PATHS.get("distancetoCoast")):
+    """
+    Computes the distance to the coastline from a given geographic point.
+
+    Args:
```
chrisjwin: please adapt docstring to the style used elsewhere in RK

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    baseRotorDiam=None,
+    defaultOffshoreParamsFp=None,
+):
+    """
+    Scales a generic offshore CAPEX value based on water depth and distance to shore.
+
+    The function splits the total input CAPEX into major cost components (turbine, foundation,
+    cable, overhead) and scales each individually based on project-specific parameters.
+
+    Args:
```
chrisjwin: please adapt docstring to the style used elsewhere in RK

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        latitude (float): Latitude in decimal degrees.
+        longitude (float): Longitude in decimal degrees.
+        waterDepthFolderPath (str, optional): Path to the folder containing water depth .tif files.
+
+    Returns:
+        float: Water depth at the specified location in meters (positive value). Returns None if not found.
+    """
+    depthFiles = glob.glob(os.path.join(waterDepthFolderPath, "*.tif"))
+    resultDepth = getRasterValueFromTifs(depthFiles, latitude, longitude)
```
chrisjwin: This can now be replaced by a direct call to gk.raster.interpolateValues(source=[tif1, tif2, ...], points=...), I have implemented multiple source extraction in gk.raster.extractValues(), see here: https://github.com/FZJ-IEK3-VSA/geokit/blob/6bc182b9bfbde7205278e26a5ae84a20442d9c7b/geokit/core/raster.py#L866

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+            continue
+    return None
+
+
+# %%
+def getRatedCostFromWaterDepth(depth, maxMonopileDepth=25, maxJacketDepth=55):
+    """
+    Estimates the rated cost of offshore wind turbine foundations based on water depth.
+
+    Args:
```
chrisjwin: please adapt docstring to the style used elsewhere in RK

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+    return cableCost
+
+
+def onshoreTcc(cp, hh, rd, gdpEscalator=None, bladeMaterialEscalator=None, blades=None):
+    """
+    Calculates the turbine capital cost (TCC) of a 3-blade onshore wind turbine based on
+    capacity, hub height, and rotor diameter according to the model by Fingersh et al.
+
+    Args:
```
chrisjwin: please adapt docstring to the style used elsewhere in RK

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    if baseCap is None:
+        baseCap = params.base_capacity
+        print("baseCap is taken from overall techno-economic file")
+    if baseHubHeight is None:
+        baseHubHeight = params.base_hub_height
+        print("baseHubHeight is taken from overall techno-economic file")
+    if baseRotorDiam is None:
+        baseRotorDiam = params.base_rotor_diam
+        print("baseRotorDiam is taken from overall techno-economic file")
+    if inputCapex is None:
```
chrisjwin: For consistency, the inputCapex should be called baseCapex to make clear that it refers to the baseline turbine just as baseCap, baseHubHeight etc

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        inputCapex (float): Reference CAPEX per kW (€/kW) that should be scaled.
+        capacity (float): Turbine rated capacity in kW.
+        hubHeight (float): Hub height in meters.
+        waterDepth (float): Site-specific water depth in meters.
+        coastDistance (float): Distance from site to nearest coast in kilometers.
+        rotorDiam (float): Rotor diameter in meters.
+        techYear (int, optional): Year of the applied technology. Default is 2050.
+        shareTurb (float, optional): Share of turbine cost in total CAPEX. Default is 0.449.
+        shareFound (float, optional): Share of foundation cost. Default is 0.204.
+        shareCable (float, optional): Share of cable/connection cost. Default is 0.181.
```
chrisjwin: I suggest to add an "Share of turbine cost in total CAPEX **in the baseline turbine reference case.** Default is 0.449." The same goes for shareTurb, shareFound, shareCable and shareOverhead, for the latter 3 also the CAPEX should be mentioned similar to shareTurb

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    cable, overhead) and scales each individually based on project-specific parameters.
+
+    Args:
+        inputCapex (float): Reference CAPEX per kW (€/kW) that should be scaled.
+        capacity (float): Turbine rated capacity in kW.
+        hubHeight (float): Hub height in meters.
+        waterDepth (float): Site-specific water depth in meters.
+        coastDistance (float): Distance from site to nearest coast in kilometers.
+        rotorDiam (float): Rotor diameter in meters.
+        techYear (int, optional): Year of the applied technology. Default is 2050.
```
chrisjwin: Add the information that this becomes relevant only if no baseCap, baseHubHeight, baseRotorDiam or baseCapex is given

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        shareTurb (float, optional): Share of turbine cost in total CAPEX. Default is 0.449.
+        shareFound (float, optional): Share of foundation cost. Default is 0.204.
+        shareCable (float, optional): Share of cable/connection cost. Default is 0.181.
+        shareOverhead (float, optional): Share of overhead/miscellaneous costs. Default is 0.166.
+        maxMonopileDepth (float, optional): Maximum depth for monopile foundations. Default is 25.
+        maxJacketDepth (float, optional): Maximum depth for jacket foundations. Default is 55.
+        litValueAvgDepth (float, optional): Reference depth in CAPEX literature. Default is 17.
```
chrisjwin: This is actually the "baseDepth" is it not? I would call it as such to make more obvious "what belongs to what". Then mention in the arg description here that the default value is based on a literature review in 2025. The same goes for below "baseDistCoast".

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        litValueAvgDistCoast (float, optional): Reference coast distance. Default is 27.
+        baseCap (float, optional): Reference turbine capacity. Loaded from CSV if not provided.
+        baseHubHeight (float, optional): Reference hub height. Loaded from CSV if not provided.
+        baseRotorDiam (float, optional): Reference rotor diameter. Loaded from CSV if not provided.
+        defaultOffshoreParamsFp (str, optional): Filepath to offshore turbine parameters CSV.
+
+    Returns:
+        float: Adjusted offshore wind CAPEX in €/kW for the given configuration.
```
chrisjwin: Just a thought: As it is effectively a scaling, this would always be in the unit of the inputCapex (better baseCapex), would it not? Maybe we remove the units here altogether and mention that output is in the cost unit of the input baseCapex. Also remove the unit in input baseCapex description above then. Does it have to be .../kW though or would the function work also for .../MW? If only the first, than this should be mentioned in baseCapex description

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        shareCable (float, optional): Share of cable/connection cost. Default is 0.181.
+        shareOverhead (float, optional): Share of overhead/miscellaneous costs. Default is 0.166.
+        maxMonopileDepth (float, optional): Maximum depth for monopile foundations. Default is 25.
+        maxJacketDepth (float, optional): Maximum depth for jacket foundations. Default is 55.
+        litValueAvgDepth (float, optional): Reference depth in CAPEX literature. Default is 17.
+        litValueAvgDistCoast (float, optional): Reference coast distance. Default is 27.
+        baseCap (float, optional): Reference turbine capacity. Loaded from CSV if not provided.
+        baseHubHeight (float, optional): Reference hub height. Loaded from CSV if not provided.
+        baseRotorDiam (float, optional): Reference rotor diameter. Loaded from CSV if not provided.
+        defaultOffshoreParamsFp (str, optional): Filepath to offshore turbine parameters CSV.
```
chrisjwin: This is needed (and takes effect!) only if no baseCap, baseHubHeight, baseRotorDiam or baseCapex is given, should be mentioned

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+    assert (
+        55 <= maxJacketDepth < 100
+    ), "Maximum depth for jacket foundation must be between 55 and 100 m"
+
+    assert (
+        maxMonopileDepth < maxJacketDepth
+    ), "Jacket depth must be greater than monopile depth"
+
+    params = OffshoreParameters(fp=defaultOffshoreParamsFp, year=techYear)
```
chrisjwin: maybe put this into an if clause: 
```python
if any([_arg is None for _arg in [baseCap, baseHubHeight, baseRotorDiam, baseCapex]):
    params = OffshoreParameters(fp=defaultOffshoreParamsFp, year=techYear)
elif not all([_arg is None for _arg in [defaultOffshoreParamsFp, techYear]):
    raise ValueError(f"techYear and defaultOffshoreParamsFp are expected to be None if baseCap, baseHubHeight, baseRotorDiam and baseCapex are provided explicitly.")
```

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    # Combine all costs
+    totalOffshoreCapex = (
+        newTurbineCost + newFoundationCost + newCableCost + overheadCostBase
+    )
+
+    return totalOffshoreCapex
+
+
+# %%
+def getRasterValueFromTifs(tiffPaths, latitude, longitude):
```
chrisjwin: This function will not be necessary anymore with the enhanced geokit.raster.interpolateValues() which can interpret multiple sources at once via an update of geokit,raster.extractValues()

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+def getRatedCostFromWaterDepth(depth, maxMonopileDepth=25, maxJacketDepth=55):
+    """
+    Estimates the rated cost of offshore wind turbine foundations based on water depth.
+
+    Args:
+        depth (float): Water depth at the installation site (in meters).
+        allowNegative (bool): Whether negative values are allowed. If False, raises ValueError if depth < 0.
+
+    Returns:
+        float: Rated cost in €/kW.
```
chrisjwin: Since this returns an absolute cost value, the currency year should also be indicated

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+    return c1 * depth**2 + c2 * depth + c3 * 1000
+
+
+# %%
+def getCableCost(distance, capacity, variableCostFactor=1.35, fixedCost=0):
+    """
+    Calculates the cost for connecting an offshore wind power plant to the coastline.
+
+    Args:
```
chrisjwin: please adapt docstring to the style used elsewhere in RK

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        distance (float): Distance to coastline in kilometers.
+        capacity (float): Power plant's capacity in kW.
+        variableCostFactor (float, optional): Cost multiplier in €/MW/km (default = 1.35).
```
chrisjwin: Does not necessarily have to be in EUR. So maybe adapt to sth. like: "Cost multiplier in currency unit per MW and km. Defaults to1.35 [€/(MW*km)]."

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        fixedCost (float, optional): Fixed connection cost (default = 0).
```
chrisjwin: Fixed connection cost in the respective currency unit. Defaults to 0 [€]

reskit/wind/economic/scale_Offshore_Capex.py
chrisjwin: I see Philipps point but would still keep this open since it is not restricted to EUR. It can be any currency unit per MW as specified above. But is this really in [currency unit]/(MW * km)? Does this actually depend on the km to coast still, here? And is the output per MW? If so, maybe scale to kW since this is the quasi-standard in all other functions?

reskit/wind/economic/scale_Offshore_Capex.py
chrisjwin: but why was this copied? Does this mean duplicated? Why can you not use the existing onshore function?

test/03_wind/economic/test_scale_Offshore_Capex.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_getCableCost():
+    short = getCableCost(10, 14000, variableCostFactor=1.35, fixedCost=0)
+    long = getCableCost(50, 14000, variableCostFactor=1.35, fixedCost=0)
+    small = getCableCost(10, 10000, variableCostFactor=1.35, fixedCost=0)
+    large = getCableCost(10, 14000, variableCostFactor=1.35, fixedCost=0)
+
+    assert short < long, "equaiton is wrong in cable cost calculations"
+    assert small < large, "equaiton is wrong in cable cost calculations"
```
chrisjwin: can this not be asserted via a sample case and np.close to make sure the output remains the same always? imagine the units were scaled by 1000 accidentally, then these tests would still pass

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+from reskit.parameters.parameters import OffshoreParameters
+
+
+# %%
+
+
+def waterDepthFromLocation(
+    latitude,
+    longitude,
+    waterDepthFolderPath=DEFAULT_PATHS.get("waterdepthFile"),
```
chrisjwin: This will be very confusing in the function overview, rather implement this via None here and then load the default path when None later, similar to how you did it with the OffshoreParams in the main function

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    resultDepth = getRasterValueFromTifs(depthFiles, latitude, longitude)
+
+    return abs(resultDepth) if resultDepth is not None else None
+
+
+# %% function to calculate the distance to the coastline
+# if you want to execute the distance to coastline more often, please separete the loading of the taserband to increase execution time
+
+
+def distanceToCoastline(latitude, longitude, path=DEFAULT_PATHS.get("distancetoCoast")):
```
chrisjwin: same here for the "DEFAULT_PATHS.get("distancetoCoast")" - set None as default and load inside function

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    Computes the distance to the coastline from a given geographic point.
+
+    Args:
+        latitude (float): Latitude in decimal degrees.
+        longitude (float): Longitude in decimal degrees.
+        path (str, optional): File path to the distance-to-coast raster.
+
+
+    Returns:
+        float or None: Distance in kilometers, or None if point is out of bounds or an error occurs.
```
chrisjwin: Distance from coast in unit of the input raster (by default 'path' in [km]), or None if point is out of bounds or an error occurs.

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+# %%
+def calculateOffshoreCapex(
+    inputCapex,
+    capacity,
+    hubHeight,
+    waterDepth,
+    coastDistance,
+    rotorDiam,
+    techYear=2050,
```
chrisjwin: move techYear to the bottom together with defaultOffshoreParamsFp as it is only needed together, and only if the base args are not given

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    defaultOffshoreParamsFp=None,
+):
+    """
+    Scales a generic offshore CAPEX value based on water depth and distance to shore.
+
+    The function splits the total input CAPEX into major cost components (turbine, foundation,
+    cable, overhead) and scales each individually based on project-specific parameters.
+
+    Args:
+        inputCapex (float): Reference CAPEX per kW (€/kW) that should be scaled.
```
chrisjwin: remove "(€/kW)" since it could be anything, also USD or whatever? only per kW is important but that is mentioned

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        waterDepth (float): Site-specific water depth in meters.
+        coastDistance (float): Distance from site to nearest coast in kilometers.
+        rotorDiam (float): Rotor diameter in meters.
+        techYear (int, optional): Year of the applied technology. Default is 2050.
+        shareTurb (float, optional): Share of turbine cost in total CAPEX. Default is 0.449.
+        shareFound (float, optional): Share of foundation cost. Default is 0.204.
+        shareCable (float, optional): Share of cable/connection cost. Default is 0.181.
+        shareOverhead (float, optional): Share of overhead/miscellaneous costs. Default is 0.166.
+        maxMonopileDepth (float, optional): Maximum depth for monopile foundations. Default is 25.
+        maxJacketDepth (float, optional): Maximum depth for jacket foundations. Default is 55.
```
chrisjwin: Can you provide a reference where the current maxJacketDepth and maxMonopileDepth come from?

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    )
+    turbineCostReference = onshoreTcc(
+        baseCap,
+        baseHubHeight,
+        baseRotorDiam,
+        gdpEscalator=1,
+        bladeMaterialEscalator=1,
+        blades=3,
+    )
+    costRatioTurbine = turbineCostNew / turbineCostReference
```
chrisjwin: Can we rename the variables here in the name of consistency to make this easier to understand? The difference between turbineCostNew and newTurbineCost below is very difficult to get. Also, we had "base" instead of "reference" in this function. I like the naming style that you used below for depth though, so maybe like this? Note that i also suggest to name the factor "scalingFactorTurbine" as this is exactly what it does.
scalingFactorTurbine = turbinePlantCost  / turbineBaseCost

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    newTurbineCost = turbineCostBase * costRatioTurbine
+
+    # Scale foundation cost
+    depthBaseCost = getRatedCostFromWaterDepth(
+        litValueAvgDepth, maxMonopileDepth, maxJacketDepth
+    )
+    depthPlantCost = getRatedCostFromWaterDepth(
+        waterDepth, maxMonopileDepth, maxJacketDepth
+    )
+    costRatioFoundation = depthPlantCost / depthBaseCost
```
chrisjwin: scalingFactorFoundation

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    newFoundationCost = foundCostBase * costRatioFoundation
+
+    # Scale cable cost
+    cableRatio = getCableCost(coastDistance, capacity) / getCableCost(
```
chrisjwin: scalingFactorCable

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+        litValueAvgDistCoast, baseCap
+    )
+    newCableCost = cableCostBase * cableRatio
+
+    # Combine all costs
+    totalOffshoreCapex = (
+        newTurbineCost + newFoundationCost + newCableCost + overheadCostBase
```
chrisjwin: Assume the turbine becomes 2x as expensive, that would not have impact on the overheadCost in our current function., What do you think about calculating the "newOverheadCost" relatively like this: first calculate `offshoreCapexNoOverhead = newTurbineCost + newFoundationCost + newCableCost`, then calculate newOverheadCost as `newOverheadCost = offshoreCapexNoOverhead * ( shareOverhead / (1 - shareOverhead) )` and last do the totalOffshoreCapex similar to how you did it, but with the "newOverheadCost" instead of base: `totalOffshoreCapex = newTurbineCost + newFoundationCost + newCableCost + newOverheadCost`. What do you think? @mstargardt

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+        return value
+
+    except Exception as e:
+        print(f"Error at Lat: {latitude}, Lon: {longitude}: {e}")
+    return None
+
+
+# %%
+def calculateOffshoreCapex(
```
chrisjwin: this function should be moved into the "offshore_cost_model.py"

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+from reskit.default_paths import DEFAULT_PATHS
+from reskit.parameters.parameters import OffshoreParameters
+
+
+# %%
+
+
+def waterDepthFromLocation(
```
chrisjwin: I suggest to move this function (and the same goes for the "distanceToCoastline()" function below) to a new file "local_values.py" in reskit/utils/. a) it is not directly cost related and might be used for other design function in the future and b) it might be of interest even to other modules in reskit in the future, think of floating PV or wave energy. What do you think @phil-fzj ?

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+            else:
+                print(f"Got NaN from {tifPath}, continuing...")
+        except Exception as e:
+            print(f"Error reading from {tifPath}: {e}")
+            continue
+    return None
+
+
+# %%
+def getRatedCostFromWaterDepth(depth, maxMonopileDepth=25, maxJacketDepth=55):
```
chrisjwin: this function should be moved into the "offshore_cost_model.py"

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,362 @@
...
+    elif depth <= maxJacketDepth:
+        c1, c2, c3 = 103, -2043, 478
+    else:
+        c1, c2, c3 = 0, 697, 1223
+
+    return c1 * depth**2 + c2 * depth + c3 * 1000
+
+
+# %%
+def getCableCost(distance, capacity, variableCostFactor=1.35, fixedCost=0):
```
chrisjwin: this function should be moved into the "offshore_cost_model.py"

test/03_wind/economic/test_scale_Offshore_Capex.py
chrisjwin, line 1: would have to be renamed when above functions are all moved, then scale_Offshore_Capex.py would not exist anymore

chrisjwin (changes requested):
Very, very nice addition to the RESkit body, thanks @mstargardt !! Only one general question left, also to @phil-fzj - would we discontinue the "offshore_turbine_capex" function and related functions in offshore_cost_model.py then? They are very nice, but helplessly outdated and cannot be updated without a new, consistent datasource as far as I understand @mstargardt ?

reskit/wind/economic/scale_Offshore_Capex.py
mstargardt: because i have to slightly adapt the old function since it loads other values as well

test/03_wind/economic/test_scale_Offshore_Capex.py
```diff
@@ -0,0 +1,64 @@
...
+        lat, lon, waterDepthFolderPath=DEFAULT_PATHS.get("waterdepthFile")
+    )
+    assert np.isclose(
+        depth, waterdepth_exact
+    ), "the waterdepthfile is not working correct"
+
+
+def test_calculateOffshoreCapex():
+    comparedCAPEX = 2943.80819956
+    calculatedCAPEX = calculateOffshoreCapex(
```
chrisjwin: /fast/home/c-winkler/libs/RESKit/test/03_wind/economic/test_scale_Offshore_Capex.py::test_calculateOffshoreCapex failed: def test_calculateOffshoreCapex():
        comparedCAPEX = 2943.80819956
>       calculatedCAPEX = calculateOffshoreCapex(
            inputCapex=3000,
            capacity=14000,
            hubHeight=150,
            waterDepth=25,
            coastDistance=25,
            rotorDiam=230,
            techYear=2050,
            shareTurb=0.449,
            shareFound=0.204,
            shareCable=0.181,
            shareOverhead=0.166,
            maxMonopileDepth=25,
            maxJacketDepth=55,
            litValueAvgDepth=17,
            litValueAvgDistCoast=27,
            baseCap=13000,
            baseHubHeight=150,
            baseRotorDiam=250,
            defaultOffshoreParamsFp=None,
        )
E       TypeError: calculateOffshoreCapex() got an unexpected keyword argument 'inputCapex'

test/03_wind/economic/test_scale_Offshore_Capex.py
```diff
@@ -0,0 +1,64 @@
...
+import numpy as np
+from reskit.default_paths import DEFAULT_PATHS
+from reskit.parameters.parameters import OffshoreParameters
+
+# GPS Coordiantes for location in North Sea
+lon = 6.2160
+lat = 53.7170
+
+
+def test_waterDepthFromLocation():
```
chrisjwin: /fast/home/c-winkler/libs/RESKit/test/03_wind/economic/test_scale_Offshore_Capex.py::test_waterDepthFromLocation failed: def test_waterDepthFromLocation():
        waterdepth_exact = 22
>       depth = waterDepthFromLocation(
            lat, lon, waterDepthFolderPath=DEFAULT_PATHS.get("waterdepthFile")
        )

test/03_wind/economic/test_scale_Offshore_Capex.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

latitude = 53.717, longitude = 6.216, waterDepthFolderPath = None

    def waterDepthFromLocation(
        latitude,
        longitude,
        waterDepthFolderPath=None,
    ):
        """
        Returns the water depth (in meters) at a given geographic location.
    
        Parameters
        ----------
        latitude : float
            Latitude in decimal degrees.
        longitude : float
            Longitude in decimal degrees.
        waterDepthFolderPath : str, optional
            Path to the folder containing water depth .tif files. Loaded from defaults if not specified.
    
        Returns
        -------
        float or None
            Water depth at the specified location in meters (always positive). Returns None if not found.
        """
    
        if waterDepthFolderPath is None:
            waterDepthFolderPath = DEFAULT_PATHS.get("waterdepthFile")
            if waterDepthFolderPath is None:
>               raise ValueError("No waterDepthFilePath is given. Please add it to default_path.yaml.")
E               ValueError: No waterDepthFilePath is given. Please add it to default_path.yaml.

reskit/wind/economic/scale_Offshore_Capex.py:43: ValueError

chrisjwin (changes requested):
some tests are failing, please ensure that the test pipeline works at least locally until the github version is adapted

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,411 @@
...
+from reskit.parameters.parameters import OffshoreParameters
+
+
+# %%
+
+
+def waterDepthFromLocation(
+    latitude,
+    longitude,
+    waterDepthFolderPath=None,
```
chrisjwin: rename to waterDepthFilePath and explicitly allow '*' spacers in the filepath (explain in docstring). The problem is that the current solution would not work for a single .tif file in a folder with many other .tif files

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,411 @@
...
+        Water depth at the specified location in meters (always positive). Returns None if not found.
+    """
+
+    if waterDepthFolderPath is None:
+        waterDepthFolderPath = DEFAULT_PATHS.get("waterdepthFile")
+        if waterDepthFolderPath is None:
+            raise ValueError("No waterDepthFilePath is given. Please add it to default_path.yaml.")
+
+
+    depthFiles = glob.glob(os.path.join(waterDepthFolderPath, "*.tif"))
```
chrisjwin: would then simply be:
```
depthFiles = glob.glob(waterDepthFilePath)
```

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,411 @@
...
+        Path to the folder containing water depth .tif files. Loaded from defaults if not specified.
+
+    Returns
+    -------
+    float or None
+        Water depth at the specified location in meters (always positive). Returns None if not found.
+    """
+
+    if waterDepthFolderPath is None:
+        waterDepthFolderPath = DEFAULT_PATHS.get("waterdepthFile")
```
chrisjwin: Add a check here if the key exists, else one has a very uninformative error. Sth like
```
if not "waterDepthFilePath" in DEFAULT_PATHS:
    raise KeyError(f"Add 'waterDepthFilePath' key with filepath value to default_paths.yaml") 
```

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,411 @@
...
+    path : str, optional
+        File path to the distance-to-coast raster. Loaded from defaults if not specified.
+
+    Returns
+    -------
+    float or None
+        Distance in kilometers, or None if the point is out of bounds or an error occurs.
+    """
+    if path is None:
+        path = DEFAULT_PATHS.get("distancetoCoast")
```
chrisjwin: Add a check here if the key exists, else one has a very uninformative error. Sth like
```
if not "distancetoCoastFilePath" in DEFAULT_PATHS:
    raise KeyError(f"Add 'distancetoCoastFilePath' key with filepath value to default_paths.yaml")

File: reskit/wind/economic/scale_Offshore_Capex.py [hunk 48]
```diff
@@ -0,0 +1,411 @@
...
+    resultDepth = gk.raster.interpolateValues(source=[depthFiles],points=(longitude,latitude))
+
+    return abs(resultDepth) if resultDepth is not None else None
+
+
+# %% function to calculate the distance to the coastline
+# if you want to execute the distance to coastline more often, please separete the loading of the taserband to increase execution time
+
+
+def distanceToCoastline(latitude, longitude, path=None):
```
chrisjwin: please rename path in the interest of readability for user and also consistency with above depth function, name could be "distancetoCoastFilePath"

chrisjwin (changes requested):

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,411 @@
...
+    """
+
+    if waterDepthFolderPath is None:
+        waterDepthFolderPath = DEFAULT_PATHS.get("waterdepthFile")
+        if waterDepthFolderPath is None:
+            raise ValueError("No waterDepthFilePath is given. Please add it to default_path.yaml.")
+
+
+    depthFiles = glob.glob(os.path.join(waterDepthFolderPath, "*.tif"))
+    resultDepth = gk.raster.interpolateValues(source=[depthFiles],points=(longitude,latitude))
```
chrisjwin: depthFiles is a list already (glob output) and [depthFiles] creates a list-in-list. So [] must be removed, like below:
```
resultDepth = gk.raster.interpolateValues(source=depthFiles,points=(longitude,latitude))
```
Also non-list source arguments are accepted by the way, [] is not even needed for string arguments

chrisjwin (changes requested):

reskit/wind/economic/scale_Offshore_Capex.py
julian-belina: The pipeline is up and running on the dev branch

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,411 @@
...
+
+    return abs(resultDepth) if resultDepth is not None else None
```
chrisjwin: This is a bit dangerous since e.g. GEBCO contains the elevation over land as POSITIVE values, so depth should be zero in such cases. Maybe introduce an optional flag "consider_only" which can be False (normal case, as is) or "negative" (in such cases, a value greater zero of say resultDepth=7.0 would be returned as 0.0, should be used for the standard GEBCO data) or "positive" in case of an inverted definition of the depth raster where depthis indicated in positive numbers

chrisjwin (changes requested):

reskit/wind/economic/scale_Offshore_Capex.py
```diff
@@ -0,0 +1,411 @@
...
+    elif depth <= maxJacketDepth:
+        c1, c2, c3 = 103, -2043, 478
+    else:
+        c1, c2, c3 = 0, 697, 1223
+
+    return c1 * depth**2 + c2 * depth + c3 * 1000
+
+
+# %%
+def getCableCost(distance, capacity, variableCostFactor=1350, fixedCost=0):
```
chrisjwin: The variableCostFactor is most probably wrong by a factor of 1000. This is a fault of Rogeau et al. but when comparing the original sources that they refer to, it should rather be 1.35. Please have a close look as well and confirm before changing this

chrisjwin (changes requested):
Substantial error in scaling, not the author's fault but probably wrong in source paper

julian-belina:
I would propose ignoring the linting error and introducing linting in a new pull request.

phil-fzj:
closes #156

reskit/wind/economic/scale_Offshore_Capex.py
mstargardt: Like the idea and will implement it @chrisjwin

mstargardt:
added new commits. last test run needs to be done. 
in theorie, everything should be fine.

mstargardt:
@chrisjwin @phil-fzj  Please check the changes :)

reskit/wind/economic/scale_Offshore_Capex.py
mstargardt: units are given in DocString. Within the code the unit is relative.

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,796 +1,561 @@
...
-    [8] Smart, G., Smith, A., Warner, E., Sperstad, I. B., Prinsen, B., & Lacal-Arantegui, R. (2016). IEA Wind Task 26: Offshore Wind Farm Baseline Documentation. https://doi.org/10.2172/1259255
-    [9] RPG CABLES, & KEC International limited. (n.d.). EXTRA HIGH VOLTAGE cables. RPG CABLES. www.rpgcables.com/images/product/EHV-catalogue.pdf
+    float
+        Rated cost in €_2023/kW.
 
+    References
+    ----------
+    Rogeau et al. (2023), Renewable and Sustainable Energy Reviews.
     """
+    depth = abs(depth)
```
chrisjwin: This can be very dangerous since it does not differ between negative and positive values, this would yield depth-based results if we actually pass elevation values. Can be critical especially along a cliff coastline. I suggest to add a default flag `negative_depths=True` and then set all values >0 to 0 (or <0 is negative_depths=False)? One COULD even make sure that values must only be <=0 when negative_depth, but I fear this is not practicable with the low resolution of the depth raster, we will sometimes have slightly positive depth (elevation above sea) near the coast... An option to deal with this would be to combine the set to zero function above with a sanity check like `assert depth<10 if negative_depth else depth>-10, f"unrealistic depth value: {depth} m` to make sure the values are somewhat realistic

mstargardt:
We should incorporate the second part of Rougeau et al. into the approach, namely the installation costs.

I will discuss this with Drin and his master student tomorrow (14.11.2025) to see what work has already been done in this direction.

If nothing substantial has been developed yet, we can set up the installation-cost component in the next iteration. However, this will require additional research and more detailed knowledge of the maritime sector (datasets, ship databases, etc.) to ensure a robust default case in the tool.

Which likely means that this task will end up with me. 😉

mstargardt:
Installation costs are now integrated in theory. However, we still need to calibrate the input data.

At the moment, the values are adapted from external sources, but they appear to be inaccurate. This will require additional work, especially since we want the tool to function reliably for users who have no background in maritime infrastructure or general offshore wind farm design.

mstargardt:
tests are running through even updating to the current dev state

mstargardt:
I tidied up and  merged the dev  into the branch to ensure smooth merging back but -> seems like pipeline is not running through for the dev, so it isn't for this branch either

mstargardt:
@phil-fzj heyho, ready for review!

phil-fzj:
@mstargardt awesome! But the tests are still failing
and there are merge conflicts with the dev branch which should be resolved

mstargardt:
Hey all,

@chrisjwin  @phil-fzj @dmarmullaku
it seems that the tests for our offshore cost model are running successfully. However, the pipeline is not passing, likely due to issues with the examples (possibly related to merge conflicts on dev).

Since Drin and Chris have already worked on the examples, do you have any immediate ideas what might be causing this? If not, I’ll take a closer look and investigate the issue in more detail.

mstargardt:
@julian-belina @phil-fzj @chrisjwin PR up to date and ready to check!

reskit/util/local_values.py
```diff
@@ -0,0 +1,126 @@
...
+from reskit.default_paths import DEFAULT_PATHS
+import glob
+import os
+from typing import Optional, Literal
+
+from . import ResError
+
+
+# %%
+
```
phil-fzj, line 13: remove

reskit/util/local_values.py
```diff
@@ -0,0 +1,126 @@
+import numpy as np
+import pandas as pd
+import geokit as gk
+from reskit.default_paths import DEFAULT_PATHS
+import glob
+import os
+from typing import Optional, Literal
+
+from . import ResError
+
```
phil-fzj, line 10: please do not use relative imports.

reskit/util/local_values.py
```diff
@@ -0,0 +1,126 @@
...
+        return abs(val)  # make depth positive
+    elif consider_only == "positive":
+        # Inverted: sea depths are positive, land/invalid negative
+        if val <= 0.0:
+            return 0.0
+        return val  # already positive depth
+    else:
+        # Legacy behavior for backward compatibility
+        return val
+
```
phil-fzj, line 88: in the docstrings its stated that the legacy behaviour is abs(resultDepth). Here you do not return the absolute value.

reskit/util/local_values.py
```diff
@@ -0,0 +1,126 @@
...
+
+# %% function to calculate the distance to the coastline
```
phil-fzj, line 90: can this be removed? if its relevant information, put it into the docstring

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
+    assert (portDistance >= coastDistance).all(), "portDistance cannot be smaller than coastDistance for any location."
 
-    anchor : str, optional
-        Turbine's anchor type only applicable for floating foundation types, by default as recommended by [1].
-        Arguments accepted are "dea" (drag embedment anchor) or "spa" (suction pile anchor).
+    # convert all other location-specific parameters to a np array for vectorized handling in subfunctions
+    capacity = np.atleast_1d(capacity)
+    hubHeight = np.atleast_1d(hubHeight)
+    rotorDiam = np.atleast_1d(rotorDiam)
+    # voltageType = voltageType
```
phil-fzj, line 98: remove?

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
+        / onshore_tcc(
+            cp=capacityRogeau,
+            hh=hubheightRogeau,
+            rd=rotordiamRogeau,
+            gdp_escalator=1,
+            blade_material_escalator=1,
+            blades=3,
+        )
+        / capacityRogeau
+    )
```
phil-fzj, line 164: to prevent unnecessary long functions, can we move this into a separate function?

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
+            B = np.array([[7.5], [18.5]])
+            C = np.array([[5.0], [30.0]])
+            D = np.array([[0.0], [90.0]])
+            E = np.array([[2.5], [40.0]])
+            # then apply the function to each "column" of the params separately and then add up
+            instCost[~isFixed] = (((1 / A) * (2.0 * _portDistance[~isFixed][None, :] / B + C) + D) * (E / 24.0)).sum(
+                axis=0
+            )
+        # make installation cost specific and convert to EUR/kW, Rogeau et al provide cost in kEUR2022 [1]
+        return (instCost * 1000 / capacity).reshape(_depth.shape)
```
phil-fzj, line 201: can we move this outside of the function to make the main function more readable?

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
+    baseWFSize : int, optional
+        The average wind farm size in [kW], by default 106858 (based on average
+        extracted from processed theWindPower.net database v2025/07).
+    baseCap : float, optional
+        Reference turbine capacity in [kW]. Loaded from CSV if not provided.
+    baseHubHeight : float, optional
+        Reference hub height in [m]. Loaded from CSV if not provided.
+    baseRotorDiam : float, optional
+        Reference rotor diameter in [m]. Loaded from CSV if not provided.
+    defaultOffshoreParamsFp : str, optional
```
phil-fzj, line 67: specify explicitly which CSV is meant here (i.e. defaultOffshoreParamsFp)

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
+        # make installation cost specific and convert to EUR/kW, Rogeau et al provide cost in kEUR2022 [1]
+        return (instCost * 1000 / capacity).reshape(_depth.shape)
+
+    # CALCULATE LOCATION-SPECIFIC COST BY SCALING THE INDIVIDUAL TURBINE COMPONENTS
+
+    # get the component cost contributions and total default cost as per Rogeau for the given reference (base) turbine
+    # consists of cost for a) turbine, b) foundation, c) cable connection - plus overhead
+    # the turbine cost consist of the turbine (machine with tower with design impacts based on onshore calculations, scaled to offshore Rogeau cost level) and installation
+    turbineBaseCostDefault = (
+        offshoreCorrfacRogeau
```
phil-fzj, line 209: for me when running the code, offshoreCorrfacRogeau is 2.618372922707181e-09 which seems unreasonably low.

phil-fzj: is this correct?

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
+        + connectionPlantCostDefault
+    )
+    # scale the plant specific default (Rogeau) cost to custom reference CAPEX level
+    totalPlantCostCustom = totalPlantCostDefault * customScalingFactor
+
+    return totalPlantCostCustom
+
+
+# %%
+# this function returns the complete connection cost, including cable and all required converters and platforms
```
phil-fzj, line 290: put all relevant information into the docstring

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
+        )
 
-    # depth depedance
-    if fixedType:
-        pass
+    if convention == "RogeauEtAl2023":
+        # calculate electrical powerstation cost based on equation (10) and table 6
+        RCPS = {"ac": 22.87, "dc": 102.93}  # EUR/kW
+        UCPS = {"ac": 3.1750000, "dc": 7.060000}  # EUR
+        assert sorted(RCPS.keys()) == sorted(UCPS.keys())  # make sure
```
phil-fzj, line 898: its difficult to follow if its not clear what these variables stand for

reskit/wind/economic/onshore_cost_model.py
```diff
@@ -143,21 +143,26 @@ def onshore_tcc(cp, hh, rd, gdp_escalator=None, blade_material_escalator=None, b
...
+        # extract default values from OnshoreParameters...
+        OnshoreParams = OnshoreParameters(
+            gdp_escalator=gdp_escalator,
+            blade_material_escalator=blade_material_escalator,
+            blades=blades,
+        )
+        # ... and overwrite the args with it
+        blade_material_escalator = OnshoreParams.blade_material_escalator
+        gdp_escalator = OnshoreParams.gdp_escalator
+        blades = OnshoreParams.blades
```
phil-fzj, line 157: is it okay that you overwrite all if only one of these parameters is not given?

reskit/testen.py
phil-fzj, line 1: please remove

test/03_wind/economic/test_offshore_cost_model.py
```diff
@@ -1,9 +1,264 @@
...
+        baseCap=None,
+        baseHubHeight=None,
+        baseRotorDiam=None,
+        defaultOffshoreParamsFp=None,
+        techYear=2050,
+    )
 
-def test_offshore_turbine_capex():
-    capex = offshore_turbine_capex(capacity=4200, hub_height=120, rotor_diam=136, depth=30, distance_to_shore=15)
+    assert np.isclose(c1, 3393, rtol=0.05), "Error in calculateSpecificOffshoreCapex"
```
phil-fzj, line 33: given that turbinePlantCostDefault is close to zero due to offshoreCorrfacRogeau befing close to zero, please check again if these test values are correct

test/03_wind/economic/test_offshore_cost_model.py
```diff
@@ -1,9 +1,264 @@
...
```
phil-fzj, line 32: please add an additional test that would result in no scaling at all. I.e. where the resulting specifc is are the same as baseSpecCapex.
This way the overall cost scaling correctness could be checked.

test/03_wind/economic/test_offshore_cost_model.py
```diff
@@ -1,9 +1,264 @@
...
+from reskit.parameters.parameters import OffshoreParameters
+import reskit._test.data as pkg_data
+from pathlib import Path
+from importlib.resources import files, as_file
 
-from reskit.wind.economic.offshore_cost_model import offshore_turbine_capex
 
+def test_calculateSpecificOffshoreCapex():
+    c1 = calculateSpecificOffshoreCapex(
+        baseSpecCapex=1500,
```
phil-fzj, line 13: please explain in the comments what each part of the test is supposed to test

test/03_wind/economic/test_offshore_cost_model.py
```diff
@@ -1,9 +1,264 @@
...
+    )
+    expected = np.array([1820, 1765])
+    np.testing.assert_allclose(c3, expected, rtol=0.05), "Error in calculateSpecificOffshoreCapex"
+
+
+def test_getSpecificOffshoreCableCost():
+    c1 = getSpecificOffshoreCableCost(
+        distance=1000, capacity=14000, voltageType="dc", variableCostFactor=1.35, fixedCost=0, year=2050
+    )
+    assert np.isclose(c1, 1350), "Error in getCableCostfuncion, possibly due to adapted function"
```
phil-fzj, line 84: I dont understand the error message. Its not very specific. What is "possibly due to adapted function" refering to?

test/03_wind/economic/test_offshore_cost_model.py
```diff
@@ -1,9 +1,264 @@
...
+        fixedCost=0,
+        year=2050,
+    )
+    expected = np.array([20000, 800, 60000])
+
+    (
+        np.testing.assert_allclose(c2, expected, rtol=0.05),
+        "Error in getCableCostfuncion, possibly due to adapted function",
+    )
+
```
phil-fzj, line 100: also here, please explain the purpose of each check/calculation in this test

test/03_wind/economic/test_offshore_cost_model.py
```diff
@@ -1,9 +1,264 @@
...
+    with pytest.raises(Exception):
+        getSpecificOffshorePlatformCost(
+            applicationType="ac",
+            capacity=10000,
+            waterDepth=None,
+            portDistance=100,
+            foundationType="jacket",
+            maxJacketDepth=55,
+            convention="RogeauEtAl2023",
+        )
```
phil-fzj, line 173: please explain why this should fail

phil-fzj (changes requested):
Thanks @mstargardt ! I left a few comments/requests.
Specifically, please take a look at offshoreCorrfacRogeau. In my test runs this factor was close to zero leading to turbineBaseCostDefault and turbinePlantCostDefault being close to zero. This potentially looks like an error.

reskit/util/local_values.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Parameters
+    ----------
+    latitude : int|float
+        Latitude in decimal degrees.
+    longitude : int|float
+        Longitude in decimal degrees.
+    waterDepthFilePath : str, optional
+       Path or pattern to one or more GeoTIFF water depth files.
+       Support wildcards such as '*'.
```
phil-fzj, line 32: Can we add a suggested source where the user could obtain such a file? Otherwise it will be difficult to use this function.

reskit/util/local_values.py
```diff
@@ -0,0 +1,126 @@
...
+    Computes the distance to the coastline from a given geographic point.
+
+    Parameters
+    ----------
+    latitude : float
+        Latitude in decimal degrees.
+    longitude : float
+        Longitude in decimal degrees.
+    distancetoCoastFilePath : str, optional
+        File path to the distance-to-coast raster. Loaded from defaults if not specified.
```
phil-fzj, line 105: Can we add a suggested source where the user could obtain such a file? Otherwise it will be difficult to use this function.

phil-fzj (changes requested):

reskit/wind/economic/offshore_cost_model.py
```diff
@@ -1,697 +1,930 @@
...
-    offshoreSubstationTopsideCost = offshoreSubstationTopsideMass * fabricationCostRate + topsideDesignCost
-    assemblyFactor = 1  # could not find a number...
-
-    offshoreSubstationTopsideLandAssemblyCost = (
-        switchgearCost + shuntReactorCost + mainPowerTransformerCost
-    ) * assemblyFactor
+    # check inputs
+    if np.any(capacity <= 0):
+        raise ValueError(f"capacity must be an int or float > 0 kW, or an array thereof")
+    if np.any(waterDepth <= 0):
```
chrisjwin, line 740: FYI @mstargardt In my test runs, I found cases where this leads to problems, could you change this to `if np.any(waterDepth < 0):` when you work on this PR anyways? Thanks a lot!

chrisjwin (changes requested):
Only one small change which should be implemented when working on this PR anyways FYI @phil-fzj

reskit/util/local_values.py
mstargardt: done

reskit/util/local_values.py
mstargardt: done

reskit/util/local_values.py
mstargardt: done

reskit/wind/economic/offshore_cost_model.py
mstargardt: no, since it is not usable outside the function due to the fact that is highly connected to Rougeaus approach

reskit/wind/economic/offshore_cost_model.py
mstargardt: Due to several erros, taht I foudn, this is fixed.

test/03_wind/economic/test_offshore_cost_model.py
mstargardt: not relevant anymore, deleted

reskit/wind/economic/offshore_cost_model.py
mstargardt: Difficult for us as well, which also leads to major confusion over the last few weeks since it is not stated clearly in Rogeau's literature. For this. I added additional information.

test/03_wind/economic/test_offshore_cost_model.py
mstargardt: corrected

test/03_wind/economic/test_offshore_cost_model.py
mstargardt: Since I designed this for the error search over the last few weeks anyway, I added this.

reskit/wind/economic/offshore_cost_model.py
mstargardt: done!

mstargardt:
Hey @phil-fzj  @phil-fzj @chrisjwin, I finally found (hopefully) all errors in our code and also incorporated your comments into the latest merge request :)

phil-fzj:
Hi @mstargardt, thank you for fixing the workflow.
I was a little lazy yesterday evening and let AI review your code. It raised several issues.
Especially one in getSpecificOffshoreCableCost:
https://github.com/FZJ-IEK3-VSA/RESKit/blob/197ebc00e0067d8d93a6aa59a80560675af5abf1/reskit/wind/economic/offshore_cost_model.py#L685-L692
The interpolation equation is incorrect: acCostPerKm * should be a +

Below is a breakdown:

## Bugs / Correctness Issues
###  AC cable cost interpolation uses multiplication instead of addition

In `reskit/wind/economic/offshore_cost_model.py` inside `getSpecificOffshoreCableCost`:

```python
acCostPerKm = acCostPerKmDict[yearBefore]
if yearAfter > yearBefore:
    acCostPerKm = (
        acCostPerKm
        * (acCostPerKmDict[yearAfter] - acCostPerKmDict[yearBefore])  # ← should be +
        * (year - yearBefore)
        / (yearAfter - yearBefore)
    )
```

The interpolation should use `+`, not `*`. Compare with the correct pattern used identically in
`calculateOffshoreFacRogeau`. As written, this produces near-zero costs for any interpolated AC year.

### `distanceToCoastline` uses the wrong key for lookup

In `reskit/util/local_values.py`:

```python
if not "distancetoCoastFilePath" in DEFAULT_PATHS:   # checks this key
    raise KeyError(...)
distancetoCoastFilePath = DEFAULT_PATHS.get("distancetoCoastPath")   # but reads THIS key
```

`default_paths.yaml` defines `distancetoCoastPath`, so `DEFAULT_PATHS.get("distancetoCoastPath")` would
return `null` — but the guard above uses `"distancetoCoastFilePath"` (different key), so it won't catch
the misconfiguration. The result is a silent `None` return rather than a clear error.


### `waterDepthFromLocation` silently ignores all but the first matched file

When a directory or wildcard returns multiple `.tif` files, only `candidates[0]` (alphabetically first)
is used without warning. For tiled datasets (e.g. GEBCO split into tiles), this will silently return
wrong values for locations not covered by the first tile.

## Code Quality

### Naming convention inconsistency
All new functions use `camelCase` (`calculateSpecificOffshoreCapex`, `getOffshoreTurbineFoundationCost`,
etc.) while the rest of the codebase uses `snake_case` (`onshore_turbine_capex`,
`offshore_turbine_capex`). The public API export in `__init__.py` now exposes a camelCase name.


### Wildcard import in core module
```python
from reskit.util.local_values import *
```

This imports `waterDepthFromLocation` and `distanceToCoastline` into the module namespace without them
being used by any function in the module. It also pollutes the namespace of anything that imports from
`offshore_cost_model`. Prefer explicit imports or remove entirely.

### `#%%` Jupyter cell markers in a production `.py` file

There are 6+ `# %%` markers in `offshore_cost_model.py`. These belong in notebooks, not library code.


## Test Issues

### Test assertions using silent tuple wrapping

Across many tests in `test/03_wind/economic/test_offshore_cost_model.py`:

```python
(
    np.testing.assert_allclose(c2, expected, rtol=0.05),
    "Error message",  # ← this message is never shown on failure
)
```

`np.testing.assert_allclose` raises on failure; it doesn't return anything. The string is discarded
silently. The correct form is:

```python
np.testing.assert_allclose(c2, expected, rtol=0.05, err_msg="...")
```


## Further Issues
offshore_turbine_capex was removed without deprecation. 
Also it is still referenced in the docstring of onshore_turbine_capex

reskit/testen.py
phil-fzj, line 1: also you pushed this empty file. please remove it

phil-fzj (changes requested):

chrisjwin:
fyi @mstargardt @phil-fzj I just adapted the wrong interpolation (* -> +)

dmarmullaku:
updated tests and small bug fixed in branch "233_update_WISDEM_turbine_cost" --> ready for review 

if review is ok, the branch can be merged back into this one "92 simplify and update offshore cost model"

phil-fzj:
hi @dmarmullaku, thank you. I checked https://github.com/FZJ-IEK3-VSA/RESKit/tree/233_update_WISDEM_turbine_cost
I saw that you fixed some the issues raised but not all of them:
These points from above are still open:
- waterDepthFromLocation silently ignores all but the first matched file
- #%% Jupyter cell markers in a production .py file
- Test assertions using silent tuple wrapping (still present in one of four checks in test_calculateSpecificOffshoreCapex)
- offshore_turbine_capex was removed without deprecation. But is still referenced in the docstring of onshore_turbine_capex

phil-fzj:
I fixed the tests. There were some issues because of the additional dependencies coming from the wisdem integration, which required python=3.11+
@mstargardt Whats the current status of the remaining issues that I mentioned?
(@chrisjwin FYI)

phil-fzj (approved):
Awesome. Good to see this finally ready!

Issue #177, opened 2025-08-07
Add DAC workflow to RESKit


phil-fzj:
Add the DAC workflow from @H-Wenzel to RESKit. It also relies on weather data from ERA5. This way we have a standardized output format.

PR #183, opened 2025-08-25
182 add air cooling workflow

### LINKED ISSUE #182: Add Air Cooling Workflow to RESKit
Add a workflow which generates timeseries data for air cooling systems depending on the ambient air temperature

H-Wenzel:
added air cooling & heat pump workflows - additionally corrected bug in DAC workflow

closes #182


reskit/dac/workflows/dac_workflow_manager.py
```diff
@@ -0,0 +1,329 @@
...
+            "conversion_factor_heat": "MWh_heat/t_CO2",
+            "conversion_factor_water": "t_H2O/t_CO2",
+            "CO2_output": "t_CO2",
+            "H2O_output": "t_H2O",
+            "electricity_input": "MWh_el",
+            "heat_input": "MWh_heat",
+        }
+        self.units = OrderedDict(units)
+
+    def calculate_relative_humidity(self):
```
chrisjwin: As a small remark as discussed with @phil-fzj : It might be good to outsource this very general function directly to the Weather Source class to make it available also for other modules/workflow managers who might deal with humidity.

chrisjwin (changes requested):
minor outsourcing of rel humidity suggested

reskit/dac/workflows/dac_workflow_manager.py
H-Wenzel: I have outsourced it into the utils, since there were already similar functions. See #177

chrisjwin: I wonder why this was not integrated into the weather stuff from the beginning, but the main point is that it should be outside of the DAC and available to other modules, plus it aligns with similar functions in this folder so fine with me! fyi @phil-fzj

chrisjwin (approved):

reskit/cooling_heating/workflows/cooling_heating_workflow_manager.py
```diff
@@ -0,0 +1,390 @@
...
+        # set the units of an air cooling system:
+        units = {
+            "capacity": "kW_th",
+            "capacity_factor": "-",
+            "conversion_factor_electricity": "kWh_el/kWh_th",
+            "conversion_factor_fan_electricity": "kWh_el/kWh_th",
+            "conversion_factor_pump_electricity": "kWh_el/kWh_th",
+            "electricity_input": "kWh_el",
+            "electricity_input_fan": "kWh_el",
+            "electricity_input:pump": "kWh_el",
```
phil-fzj: why is there a colon before "pump"

reskit/cooling_heating/workflows/cooling_heating_workflow_manager.py
```diff
@@ -0,0 +1,390 @@
...
+
+        # Air Cooler Cost stays the same in any case:
+        alpha = 1.135  # kW/(m2K) [2]
+        A = 1 / (alpha * heatTransferDelta)  # needed A-frame size for 1 kW of cooling
+        CAPEXAC = 2.8 * 156000 * (A / 200) ** 0.89  # [1]
+
+        CAPEXDesign = CAPEXFanDesign + CAPEXPumpDesign + CAPEXAC
+        CAPEX = CAPEXFan + CAPEXPump + CAPEXAC
+
+        self.sim_data["capacity_factor"] = CAPEXDesign / CAPEX
```
phil-fzj: i am a little confused with the definition of "capacity_factor" that you applied here.
Inside reskit and i think also in general the capacity factor is the ratio of electrical energy output (time-dependent) to the maximum electrical energy output or the nominal capacity.

i would not call this capacity factor. I see where this i coming from but i think its quite confusing to have a time-dependent capex which is used to determine a capacity factor. This needs a little more explanation

reskit/cooling_heating/workflows/cooling_heating_workflow_manager.py
```diff
@@ -0,0 +1,390 @@
...
+        self.sim_data["conversion_factor_electricity"] = (
+            -1 / self.sim_data["COP"]
+        )  # kWhel/kWhth
+
+        # set the units of an air source heat pump:
+        units = {
+            "capacity": "kW_th",
+            "conversion_factor_electricity": "kWh_el/kWh_th",
+            "electricity_input": "kWh_el",
+        }
```
phil-fzj: missing unit for COP

phil-fzj (changes requested):
minor feedback. otherwise looks good to me!

reskit/cooling_heating/workflows/cooling_heating_workflow_manager.py
```diff
@@ -0,0 +1,425 @@
...
+                    1,
+                ],
+            ]
+        ).T  # density in kg/m3
+        self.airData = pd.DataFrame(
+            index=[-30, -20, -10, 0, 10, 20, 30, 40, 50, 60, 70, 80],
+            data=airData,
+            columns=["cp", "density"],
+        )  # index referes to ambient air temperature
+
```
phil-fzj, line 87: why is this in the init? (its only used by 1 of 2 workflows)
Is this something where we expect that other workflows might need it too?

phil-fzj (commented):
looks good to me. 
Only one mior question left. Afterwards we can merge

reskit/cooling_heating/workflows/cooling_heating_workflow_manager.py
H-Wenzel: exactly. I suppose other cooling workflows which will be added in the future, like evaporative cooling, might also need this data

phil-fzj (changes requested):
looks good. ready to merge

phil-fzj (approved):
looks good. ready to merge

PR #184, opened 2025-09-01
Air cooling capacity factor

H-Wenzel:
Corrected the calculation of the capacity factor to include reductions of the capacity between the design temperature and the maximum operating temperature.


phil-fzj (approved):
looks good to me.

PR #187, opened 2025-09-08
186 scale windspeed

chrisjwin:
Allows alternative wind speed scaling from reference to hub height via long-run-average shear slopes. Also correctly represents influence of planetary boundary layer now under all cases of planetary boundary layer height constellations. Outsources several functions and adds tests.


phil-fzj (approved):
looks good

PR #190, opened 2025-10-03
update test and example and documentation for CSP

### LINKED ISSUE #118: update test and example and documentation for CSP
Tests for CSP are skipped with hardcoded paths.

Examples for CSP are not working because workflow names are now correct

check documentation

also remove hardcode in other solar examples: E.g. 4.01-Solar-Load_Module.ipynb
Image

This issue was originally created by: @phil-fzj

k-okosun:


reskit/csp/workflows/csp_workflow_manager.py
```diff
@@ -393,8 +393,8 @@ def apply_capacity_sf(self):
             ).all()
         else:
             assert (
-                self.sim_data["HeattoHTF_W"].mean(axis=0)
-                / self.placements.capacity_sf_W_th
+                self.sim_data["HeattoHTF_W"].mean(axis=0)[0]
+                / self.placements.capacity_sf_W_th.values
             ) < 1
```
phil-fzj: is self.sim_data["HeattoHTF_W"] one-dimensional?
If yes, can we do this more cleanly?

reskit/csp/workflows/workflows.py
```diff
@@ -144,26 +153,74 @@ def CSP_PTR_ERA5_specific_dataset(
...
+            Defaults to False.
+        onlynightuse (bool, optional):
+            If True, optimizes plant size for night use only.
+            Defaults to True.
+        fullvariation (bool, optional):
+            If True, applies full variation in calculations.
+            Defaults to False.
+        _validation (bool, optional):
+            If True, runs in validation mode (some input data may be replaced).
+            Defaults to False.
```
phil-fzj: you are using a different docstring format than the rest of reskit.
We use the NumPy documentation style but you seem to use the google-style.
Please adapt this so we have consistent docstring styles across reskit

reskit/csp/workflows/workflows.py
```diff
@@ -217,7 +274,8 @@ def CSP_PTR_ERA5_specific_dataset(
     # do long run averaging for DNI
     # TODO: remove
     if global_solar_atlas_dni_path == "default_cluster":
-        global_solar_atlas_dni_path = r"/storage/internal/data/gears/geography/irradiance/global_solar_atlas_v2.5/World_DNI_GISdata_LTAy_AvgDailyTotals_GlobalSolarAtlas-v2_GEOTIFF/DNI.tif"
+        global_solar_atlas_dni_path = r"/fast/central/shared_data/2023_gears/geography/irradiance/global_solar_atlas_v2.5/World_DNI_GISdata_LTAy_AvgDailyTotals_GlobalSolarAtlas-v2_GEOTIFF/DNI.tif"
+
```
phil-fzj: we cant have hardcoded paths in here.
No one without access can use this

phil-fzj: i would suggest to just remove this

phil-fzj (changes requested):

reskit/csp/workflows/csp_workflow_manager.py
phil-fzj: why use the if clause (not visible in snippet, but distinguishes between placements with len=1 and len>1) in the first place. we can just remove it and as this part works for both cases:
```python

            assert (
                (
                    self.sim_data["HeattoHTF_W"].mean(axis=0)
                    / self.placements.capacity_sf_W_th
                )
                < 1
            ).all()

phil-fzj (approved):
Awesome. Thank you Kenneth!

PR #200, opened 2025-11-18
Version used for wind workflow publication

phil-fzj:
add changes that were not pushed previously:
- add more robust handling of new ws extrapolation
- fix bug with loading power curves from folder path


phil-fzj:
adapted test values for wind workflow and wind_config due to updated ws and cf correction files

phil-fzj:
this PR also updates the version number to 0.4.3 so that we have a release on zenodo with a DOI.

reskit/wind/core/data/ws_correction_factors_PSDW2025.yaml
```diff
@@ -1,250 +1,390 @@
...
+  35.7-35.8: .nan
+  35.8-35.9: .nan
+  35.9-36.0: .nan
+  36.0-36.1: .nan
+  36.1-36.2: .nan
+  36.2-36.3: .nan
+  36.3-36.4: .nan
+  36.4-36.5: 0.294627913325717
+  36.5-36.6: .nan
+  36.6-36.7: .nan
```
julian-belina: Are nan values expected in this file?

julian-belina (approved):
Looks good to me otherwise

reskit/wind/core/data/ws_correction_factors_PSDW2025.yaml
phil-fzj: no good catch. i fixed that. the values should have stopped at 32 m/s

Issue #210, opened 2025-11-26
add workflows for evaporative cooling systems


H-Wenzel:
The idea is to expand the cooling_heating module to also enable simulation of evaporative cooling systems. The calculations for this come from @BernhardWortmann and are already existing but need to be translated to python and integrated to RESKit

H-Wenzel:
Translated evaporative cooling model from Julia to Python and integrated it into RESKit as a cooling_heating workflow.
The workflow is finished and can calculate the water demand of an evaporative cooling tower under varying weather conditions. 
Branch [#210](https://github.com/FZJ-IEK3-VSA/RESKit/tree/210-add_evaporative_cooling_wf)

- [x] Translate model from Julia to Python
- [x] Integrate WF into RESKit
- [x] Write Tests

PR #213, opened 2025-11-27
210 add evaporative cooling wf

H-Wenzel:
Added evaporative cooling system workflow of @BernhardWortmann to RESKit. 
#210


phil-fzj (approved):
Looks good!

PR #221, opened 2026-01-08
solve issue#115 make ERA5 reading compatible with u and v

### LINKED ISSUE #115: process u and v of wind speed on the fly
process u and v of wind speed on the fly within RESKit so that we don´t necessarily do a preprocessing step on the ERA5 data
This issue was originally created by: @phil-fzj

shuying-chen-fzj:
By using this new development, when users specify to read "elevated_wind_speed" in wf.read(variables), wf first searches variable ws100, if it exists, then the reading process is done. If not, wf tries to read u and v and compute wind speed to be the "elevated_wind_speed" in the data part of wf. This function works for all the wind speed variables: "elevated_wind_speed", "surface_wind_speed", "wind_speed_at_100m", "wind_speed_at_10m", i.e., first try read precomputed ws<X>m, if doesn't work then try to read u and v to be the target wind speed.

I have tested, when giving a ERA5 directory containing ws100, or ws100 + u + v, or u + v. The codes passed the test. Of course, neither ws100 nor both u100 and v100 would raise an error.


phil-fzj (changes requested):
Thanks! 
You mention that you have manually tested you code.
It would be great, if you could add an actual test for this in reskit.
E.g. one that checks if the outputs are the same when using the precomputed wind speed against the case where this is calculated from the u and v component directly in reskit:

shuying-chen-fzj:
A new test has been added to check if the loaded wind speed, calculated (on the fly) based on its u and v components, is correct at mean and an example test point values. A comparison to the other pre-calculated wind speed shows almost no difference (difference occurs at the 4th digital after the decimal, understandable given the data accuracy) when using this wind speed reading based on u and v. Please review the pull request again with new commits included.

phil-fzj (approved):
Thank you Shuying!

PR #230, opened 2026-02-19
fixed bug if execute_workflow_iteratively is executed twice

### LINKED ISSUE #212: execute_workflow_iteratively fails if run twice

- [x] I have confirmed this bug exists on the lastest commit on the master branch

- [x] I have confirmed this bug exists for a specific version the software.

### Issue Description

If one runs execute_workflow_iteratively twice, we get an 
> AssertionError: placements must not have 'RESKit_sim_order' attribute

This is because the placements dataframe passed by the user is overwritten during the code execution in RESKit.

Should we automatically remove any columns called RESKit_sim_order before running the workflow?
Otherwise we could just create a copy of it within RESKit.

### Reproducible Example

```python
import reskit as rk
import pandas as pd
import numpy as np

# Create Placements DataFrame with turbine locations and specifications

placements = pd.DataFrame(
    {
        "lon": [5.985195, 5.994685, 6.004750],
        "lat": [50.797254, 50.794208, 50.784432],
        "hub_height": [120, 120, 82],
        "capacity": [4000, 4000, 4000],
        "rotor_diam": [130, 150, 136],
        "powerCurve": [np.nan, np.nan, "V117-3300"],
    }
)
from reskit.workflow_manager import execute_workflow_iteratively

workflow_args = {
    "placements": placements,
    "era5_path": rk.TEST_DATA["era5-like"],
    "gwa_100m_path": rk.TEST_DATA["gwa100-like.tif"],
    "height_scaling_data": {
        50: rk.TEST_DATA["gwa50-like.tif"],
        200: rk.TEST_DATA["gwa200-like.tif"],
    },
}

execute_workflow_iteratively(
    rk.wind.wind_era5_PenaSanchezDunkelWinklerEtAl2025,
    "era5_path",
    zoom=4,
    **workflow_args,
)
execute_workflow_iteratively(
    rk.wind.wind_era5_PenaSanchezDunkelWinklerEtAl2025,
    "era5_path",
    zoom=4,
    **workflow_args,
)
```

### Expected Behavior

No error.

### pip list -v

```bash

```

H-Wenzel:
fixed bug by checking:
1) if "RESKit_sim_order" is already in placements, it needs to be "correct", i.e. the same as if it would be freshly set
2) if "source" already in placements, it needs to contain the correct tile paths

### DIFF
--- a/reskit/util/weather_tile.py
+++ b/reskit/util/weather_tile.py
@@ -4,6 +4,7 @@ import geokit as gk
 import osgeo
 import pandas as pd
 from smopy import deg2num
+import numpy as np
 
 
 def get_tile_XY(zoom, lon=None, lat=None, geom=None):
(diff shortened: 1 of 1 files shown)


chrisjwin (approved):
Thanks a lot!

PR #231, opened 2026-02-19
217 check if realistic turbine capacity is given by user

### LINKED ISSUE #217: Check if realistic turbine capacity is given by user

Add a check to make sure that the user provides the capacity in kW and not in another unit.

### Proposal

Add logic that calculates specific power and checks if its realistic.

H-Wenzel:
added check for realistic specific power values (i.e. W/m2, capacity per rotor area), derived from turbine library


reskit/wind/core/power_curve.py
```diff
@@ -47,7 +47,7 @@ def compute_specific_power(capacity, rotor_diam, **k):
     Returns
     -------
     float or array-like
-        Specific power in kW/m2
+        Specific power in W/m2
```
chrisjwin, line 50: Is this actually the case? I see `*1000` in line 53, to me that sounds a lot like kW as input? I may be mistaken though

chrisjwin: Also, below the assert would return an error as "Check if capacity is defined in correct unit (kW)"

chrisjwin (dismissed):
Please have a look at my comment

reskit/wind/core/power_curve.py
H-Wenzel: Yes, the capacity input is in kW, but by multiplying it with 1000 the output becomes W/m2 rather than kW/m2

chrisjwin: my bad, I was under the impression this was the input unit but it is the output, so all OK!

phil-fzj (approved):

PR #277, opened 2026-06-11
155 reskit weather data prepare

### LINKED ISSUE #155: Integrate weather data download/preparation into RESKit
This issue was originally created by: @shitabishmam

phil-fzj:
The whole workflow of downloading, processing and tiling ERA5-based workflows is implemented and callable as a reskit function:

result = rk.download_and_process(
workflow="wind_era5_PenaSanchezDunkelWinklerEtAl2025",
start_date="2000-01-01",
end_date="2000-12-31",
boundary_box={"north": 55, "south": 47, "west": 6, "east": 15}, # Germany
output_dir=output_dir,
tiling=True,
zoom_level=4, # 16×16 global tile grid
)

I.e.: Optional tiling has been implemented as well (with padding and antimeridian fixes)

Two examples exist that demonstrate usage for wind and pv:
examples/1_load_input_data/1_1_3_prepare_era5_for_wind_workflow.py
examples/1_load_input_data/1_1_4_prepare_era5_for_solar_workflow.py

For the wind_era5_PenaSanchezDunkelWinklerEtAl2025 workflow we also need GWA data which we cannot automatically download at the moment. Therefore, for GWA, there’s a separate message explaining how to download the data.

The rk.download_and_process now supports all ERA5 based reskit workflows.
If the user passes a currently not supported workflow (e.g. one depending on MERRA2 data), a NotImplementedError is raised.

If the user wants to download all necessary variables at once to be able to run all ERA5-based workflows, they can do so by selecting:
workflow = all_era5_workflows


phil-fzj:
closes #155

reskit/util/input_preparation.py
```diff
@@ -0,0 +1,199 @@
+import importlib
+import inspect
+import os
+import pkgutil
+import reskit as rk
+
+depends_on = {
+    "wind_era5_PenaSanchezDunkelWinklerEtAl2025": {
+        "GWA4": ["50m", "100m", "200m"],
```
chrisjwin: I suggest to use more specific names that indicate "windspeed_50m" etc. or so, similar to the ERA5 approach, given that we may also use other GWA variables as we already do with GSA

reskit/util/input_preparation.py
```diff
@@ -0,0 +1,199 @@
...
+            var
+            for workflow, deps in depends_on.items()
+            if workflow not in _NON_WORKFLOW_KEYS
+            for var in deps.get("ERA5", [])
+        )
+    ),
+}
+
+# Maps CDS API variable names to their short names inside the downloaded NetCDF file.
+_era5_cds_to_nc_name = {
```
chrisjwin: This should be stored centrally and hence moved to the ERA5 class (where it is dearly needed by the way in the "load" functions because it is very hard to track the actual parameter in the current loader structure)

reskit/util/input_preparation.py
```diff
@@ -0,0 +1,199 @@
...
+    "surface_pressure": "sp",
+    "boundary_layer_height": "blh",
+    "forecast_surface_roughness": "fsr",
+    "surface_solar_radiation_downwards": "ssrd",
+    "total_sky_direct_solar_radiation_at_surface": "fdir",
+}
+
+# Variables whose NC short names are consumed by preprocess_era5_data and replaced
+# by derived outputs (e.g. u100+v100 → ws100). These should not be re-tiled as raw.
+_era5_preprocessed_nc_names = {"u10", "v10", "u100", "v100", "ssrd", "fdir"}
```
chrisjwin: For consistency, this might then also be moved to the ERA5 source class ?

chrisjwin (changes requested):

reskit/util/input_preparation.py
phil-fzj: changed to ["wind-speed_50m", "wind-speed_100m", "wind-speed_200m"]

chrisjwin (approved):

PR #278, opened 2026-06-11
integrate snow data reading from ERA5 data

shuying-chen-fzj:
Era5Source is now able to read snow data (snow albedo, density, depth, and snowfall).
The corresponding test functions for reading these snow variables are also added (added example snow data in the test data).
Snow variable names are also added to the variable read list of solar workflow openfield_pv_era5.


phil-fzj (approved):

phil-fzj:
Thank you!

Issue #281, opened 2026-06-24
update to newer geokit version


phil-fzj:
### Purpose of the improvement

Accepted datadtypes for gk.raster.interpolateValues changed with the new geokit version. Therefore some tests are failing with:
> FAILED test/01_util/util/test_topography.py::test_visibility_from_topography - geokit.error.GeoKitRasterError: The following datatype has been provided as point but something different was expected: <class 'numpy.ndarray'>

### Proposal

Convert points dtypes passed to gk.raster.interpolateValues to dypyes accepted by geokit.

Currently accepted:
```
points : (X,Y) or [(X1,Y1), (X2,Y2), ...] or Location or LocationSet()
Coordinates for the points to extract
* All points must be in the same SRS
* !REMEMBER! For lat and lon coordinates, X is lon and Y is lat
  (opposite of what you may think...)
```

julian-belina:
@phil-fzj Which other formats might be required? Perhaps I could add them to GeoKit.

phil-fzj:
geokit < 1.7 allowed any kind of iterable to be passed to interpolateValues (but this was not documented in the docstrings)
```
    else:  # Assume points is already an iterable of some sort
        asSingle = False
```

In 1.7 this is more strict and limited to:
```
    points: tuple[float, float]
    | ogr.Geometry
    | list[tuple[float, float]]
    | list[ogr.Geometry]
    | Location
    | LocationSet,
```

I think its okay like this. We will just adapt the relevant code parts in reskit to always pass the allowed datatype.

PR #283, opened 2026-06-25
172 update to latest version of pvlib

### LINKED ISSUE #172: Update to latest version of pvlib
Latest version v0.13.0 as of now does not comply with RESkit code, arguments have changed. Adapt to latest version, ideally with backward compatibility

phil-fzj:
Currently pvlib is pinned to 0.9.0

In pvlib 0.10.0 values for module_dc_voltage_at_mpp change. This is due to a fix by pvlib in 0.10.0:

Prevent small negative values of v_oc in pvlib.singlediode._lambertw() which result from accumulated roundoff error. (https://github.com/pvlib/pvlib-python/issues/1780, https://github.com/pvlib/pvlib-python/issues/1673, https://github.com/pvlib/pvlib-python/pull/1782)

That fix changes the single-diode MPP search slightly and shifts v_mp enough to move the interpolated module_dc_voltage_at_mpp regression value.

Other test values in our workflows are not affected. 
As a consequence, we will need to adapt the slightly changed test values for module_dc_voltage_at_mpp:

    e.g. 37.39738372317519 -> 37.39337385334944


phil-fzj:
note that this already includes #281. so #281 should be merged first

chrisjwin (approved):

PR #287, opened 2026-07-09
214 enable workflow wind config to skip lra

### LINKED ISSUE #214: wind_config workflow does not support skipping the long run average (LRA) step

- [ ] I have confirmed this bug exists on the lastest commit on the master branch

- [ ] I have confirmed this bug exists for a specific version the software.

### Issue Description

when using the wind_config as a generic wind power workflow, it should be possible to skip the LRA step as it's not applicable to every weather data. At the moment, wind_config does not work out as it says it would be.

### Reproducible Example

```python
To call other weather data other than ERA5 using wind_config
```

### Expected Behavior

did not work out for me

### pip list -v

```bash

```

shuying-chen-fzj:
This is a minor patch to the wind_config workflow. Now it is able to skip the adjustment using the longterm run average, which is only useful for workflows with a need of further downscaling for the weather data in use.

### DIFF
--- a/reskit/wind/workflows/workflows.py
+++ b/reskit/wind/workflows/workflows.py
@@ -399,8 +399,8 @@ def wind_config(
     placements,
     weather_path,
     weather_source_type,
-    enable_lra_adjustment,
     weather_lra_ws_path,
+    enable_lra_adjustment,
     real_lra_ws_path,
     real_lra_ws_scaling,
     real_lra_ws_spatial_interpolation,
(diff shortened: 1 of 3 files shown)


chrisjwin:
@shuying-chen-fzj thanks and will be merged shortly, in sync with @phil-fzj I have only adapted the logic a little, it was not so straightforward to have a function `adjust_variable_to_long_run_average()` - which has an argument flag `enable_lra_adjustment = False`, basically saying "do NOT adjust variable to long run average". So we have added an if clause which will skip the whole LRA adjustment function if `enable_lra_adjustment` in the workflow `wind_config()` is False.

chrisjwin (approved):

PR #296, opened 2026-07-16
233 update wisdem turbine cost

### LINKED ISSUE #233: integrate latest WISDEM turbine cost function

Instead of the currently hardcoded parameters in onshore_tcc(), we (@phil-fzj and me) suggest to use the WISDEM package which has an updated cost function. The new function could be called inside onshore_tcc() so that nothing else is affected besides slightly changing cost values. The new version a) allows greater diameters and b) considers scale effects for very small turbines, all in all representing edge cases/extrapolations better

fyi @julian-belina 

### Proposal

Integrate https://github.com/NLRWindSystems/WISDEM into the imports and call their nrel_csm_2015() within onshore_tcc()

phil-fzj:


phil-fzj:
closed. This was just for running the full pipeline
will be merged as part of #169

PR #302, opened 2026-07-22
84 era 5 weather long term is missing for values 179179 longitude

### LINKED ISSUE #84: ERA-5 weather long term is missing for values <179°/>179° longitude
This issue was originally created by: @chrisjwin

phil-fzj:
Our old ERA5 LRA files did not have full global coverage and could not be reproduced.
Therefore, functions were added to recreate the LRA files in this PR.
New LRA files are provided with this PR.


chrisjwin (changes requested):
Great to finally have this in RK! As discussed, we should only add a versioniong for the LRA files and tests for the preprocessing functions. Thanks

chrisjwin (approved):
Great and thanks a lot!

PR #303, opened 2026-07-23
started revision of chained code assignment

### LINKED ISSUE #219: Change code where values are set through chained assignment

Pandas will not support setting values through chained assignment in pandas 3.0 anymore.
Currently, when using pandas 2, this leads to warnings.

```python
NCSource.py:258: FutureWarning: ChainedAssignmentError: behaviour will change in pandas 3.0!
You are setting values through chained assignment. Currently this works in certain cases, but when using Copy-on-Write (which will become the default behaviour in pandas 3.0) this will never work to update the original DataFrame or Series, because the intermediate object on which we are setting values will behave as a copy.
A typical example is when you are setting values in a column of a DataFrame, like:
 
df["col"][row_indexer] = value
 
Use `df.loc[row_indexer, "col"] = values` instead, to perform the assignment in a single step and ensure this keeps updating the original `df`.
 
See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
 
  self.variables["path"][time_name] = self.variables["path"][time_index_from]
```

### Proposal

Change all occurrences in the code where values are set through chained code assignment.
Check if there are no more warnings.

chrisjwin:
adapts pandas indexing to new pandas version to avoid slicing warnings


chrisjwin (commented):
In case that this does not yet cover all occurrences of the pandas slicing issues, a new issue should be opened and the respective code lines can then be changed as needed. Good to go so we can close this for now!

phil-fzj (approved):

PR #319, opened 2026-08-20
286 update to pandas v3 yields very different outputs

### LINKED ISSUE #286: update to pandas v3 yields very different outputs

- [ ] I have confirmed this bug exists on the lastest commit on the master branch

- [ ] I have confirmed this bug exists for a specific version the software.

### Issue Description

See problem in test pipeline of pull request https://github.com/FZJ-IEK3-VSA/RESKit/pull/261 here: https://github.com/FZJ-IEK3-VSA/RESKit/actions/runs/29013360996/job/86102381797?pr=261 

It is not only a dependency clash but the output values change fundamentally (not a numeric issue, different dimensions of the dataframe), causing the asserts to fail

### Reproducible Example

```python
pipeline of pull request https://github.com/FZJ-IEK3-VSA/RESKit/pull/261 here: https://github.com/FZJ-IEK3-VSA/RESKit/actions/runs/29013360996/job/86102381797?pr=261
```

### Expected Behavior

Check what the reason is and find a way to ideally allow both pandas v2 and v3 with successful tests

### pip list -v

```bash

```

jenszimmert:
Resolves issue #286. Updating to pandas v3 does change the number of warnings in testing. However, it seems that this is not related to pandas directly but to different warnings filtering. Using '-W always' as pytest option does result in identical numbers of warnings.

The changes:
- changing the pd.Series creation to dtype="object", which was earlier done as a silent upcast
- removing positional indexing in lambda functions, as positional indexing of pd.Series is now deprecated and introducing list comprehension-based solutions that could potentially be faster
- chained assignment is no longer supported, e.g. self.variables["path"][time_name] = self.variables["path"][time_index_from]. This did however just cause a warning and not a test failure.
- numpy objects created with .values will in some cases be read-only due to the copy-on-write principles implemented in pandas 3. This could be solved through the .to_numpy() function or through a .loc implementation.
- changing freq="H" to freq="h" as the former is now deprecated
- a check using np.issubdtype did no longer work as pandas introduced own dtypes. I introduced the pandas-based is_numeric_dtype instead.
- asserts with "is None" were failing because pandas now resorts to nan != None. I changed the assert to pd.isna()

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -8,7 +8,7 @@ dependencies:
   - pip <=26.1.2
   - matplotlib <=3.10.5
   - descartes <=1.1.0
-  - pandas >=2.0.0,<=2.3.3
+  - pandas >=2.0.0,<=3.0.3
   - openpyxl <=3.1.5
   - numpy >=1.25.0,<=1.26.4
   - numba >=0.58.1,<=0.65.1
--- a/reskit/csp/workflows/csp_workflow_manager.py
+++ b/reskit/csp/workflows/csp_workflow_manager.py
@@ -392,7 +392,7 @@ class PTRWorkflowManager(SolarWorkflowManager):
             )
 
             self.sim_data["solar_zenith_degree"][:, location_iter] = np.nan_to_num(
-                _solarpos["apparent_zenith"].values, 0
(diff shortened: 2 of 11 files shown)


phil-fzj (approved):
Thank you @jenszimmert. Looks good.

PR #320, opened 2026-09-01
Several Bugfixes

phil-fzj:
This fixes several bugs that were discovered during a codebase audit.

### Fixes

- **BUG-01 — README quickstart could not run.** It called
  `download_and_process(workflow=…)`, but the public keyword is `workflows`.
  `workflow` stays as a deprecated alias.
- **BUG-02 — Placement coordinate validation was wrong.** The checks used
  `.any()`, so one valid row made an invalid batch pass. They now use `.all()`,
  raise `ValueError`, and name the offending rows. A single placement at (0, 0)
  works, because the artificial extent width grows additively instead of by
  multiplication.
- **BUG-03 — `to_xarray()` changed the workflow state.** It deleted `location_id`
  from `.placements` and appended to the caller's `output_variables` list. Both
  now use copies, so a second export gives the same result.
- **BUG-04 — Loading one result file failed.** The single-file branch sorted by
  `locations`, but the datasets use `location`. Both branches now use `sortby` and
  the given `loader`. An empty input raises a clear `ValueError`.
- **BUG-05 — Parameter CSV validation did nothing.** It compared against a category
  name which the classifier never returns, and it built an `AttributeError` without
  a raise. It now raises, and it matches the `_unit` suffix exactly.
- **BUG-06 — A cross-year daily export dropped its first day.** The deletion was
  unconditional. Removed.
- **BUG-07 — Batch-size validation accepted invalid values.** The check combined
  the type test and the positivity test with `and`, so an integer zero passed and
  later caused a division by zero. It now rejects `0`, negatives, floats, strings
  and booleans.
- **BUG-08 — `distribute_workflow()` changed the caller's DataFrame.** It replaced
  the index, added `location_id`, and could delete `geom`. It copies on entry.
- **BUG-09 — Topographic sampling used the wrong trigonometric factor.** Longitude
  offsets were divided by `sin(latitude)`. The conversion from an east-west arc
  uses `cos(latitude)`. The old form divided by zero at the equator. Reference
  values updated.
- **BUG-10 — Unsupported DAC inputs did not raise the documented error.** An
  unknown fill method and an unknown HT model used asserts, which give the wrong
  exception type and disappear under `python -O`. Both now raise
  `NotImplementedError`.
- **BUG-11 — Dependency deprecations which the suite exercises.** Removed the
  pandas chained assignment, positional `Series` access, a string written into a
  float column, the `freq="H"` spelling, xarray `drop()`, an implicit xarray
  `join`, the renamed pvlib `singleaxis` azimuth argument, and `logging.warn()`.
- **BUG-12 — Tilt conventions used an unrestricted `eval`.** Empty globals still
  give the Python built-ins to the expression, so an untrusted configuration could
  run arbitrary code. `location_to_tilt()` now accepts a callable, and a string is
  evaluated by a restricted AST walker. String expressions are deprecated.
- **`fillMethod="offTmin"` always crashed** with
  `AttributeError: 'dict' object has no attribute 'loc'`. The branch called pandas
  syntax on a dict of numpy arrays, and it had no test.

### Tests

- A README smoke test compiles every python block, checks the documented keywords
  against the public signatures, and runs the `download_and_process` example with
  stubbed weather preparers, so no provider is contacted.
- Both skipped solar-core placeholders are now real tests.
- New regression tests cover the workflow manager, the parameter csv schema, the
  wind batch size, the DAC models and fill methods, and the tilt conventions.


phil-fzj:
@chrisjwin I put all of this in one PR. If you think single PRs are easier, I can do that too.

reskit/parameters/parameters.py
```diff
@@ -110,29 +110,33 @@ def _get_value(data, year):
                 )
             params_df.set_index("year", inplace=True)
 
+            # drop the nameless columns which pandas creates from a written row index
```
chrisjwin, line 113: Maybe rather add `, ignore_index=True` to above pd.read_csv() to avoid creating the Unnamed columns in the first place?

reskit/solar/core/system_design.py
```diff
@@ -1,10 +1,107 @@
...
+        if isinstance(node, ast.Constant):
+            if isinstance(node.value, bool) or not isinstance(node.value, (int, float)):
+                raise ResError(f"Tilt convention expression must not contain the constant: {node.value!r}")
+            return node.value
+        if isinstance(node, ast.Name):
+            if node.id not in names:
+                raise ResError(f"Tilt convention expression must not use the name: '{node.id}'")
+            return names[node.id]
+        if isinstance(node, ast.UnaryOp) and type(node.op) in _TILT_UNARY_OPERATORS:
+            return _TILT_UNARY_OPERATORS[type(node.op)](_evaluate(node.operand))
```
chrisjwin: Those returns are untested yet depend on the hardcoded table above so might fail when someone changes it. We should briefly exchange how far we want to drive the whole test coverage issue

reskit/solar/core/system_design.py
```diff
@@ -47,16 +147,25 @@ def location_to_tilt(locs, convention="Ryberg2020", **kwargs):
     """
     locs = gk.LocationSet(locs)
 
-    if convention == "Ryberg2020":
+    if callable(convention):
+        tilt = convention(locs.lats)
```
chrisjwin: Where does the callable option come from?

reskit/util/input_preparation.py
```diff
@@ -203,15 +205,54 @@ def _prepare_gsa(variables, **_):
 #######################################################
 
 
+def _resolve_workflows_argument(workflows, workflow):
```
chrisjwin: Should the default for the deprecated workflow arg not be None then?

reskit/util/input_preparation.py
```diff
@@ -203,15 +205,54 @@ def _prepare_gsa(variables, **_):
...
+
+    ``download_and_process()`` takes ``workflows``. Older documentation used the singular
+    ``workflow``. That spelling still works, but it gives a ``DeprecationWarning`` and it
+    will be removed.
+
+    Parameters
+    ----------
+    workflows : str or list of str or None
+        The value of the ``workflows`` argument.
+    workflow : str or list of str or None
```
chrisjwin: Should we even list it in the Docstring then? Else we can maybe more clearly flag it as deprecated? Sth along these lines https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated ?

reskit/util/input_preparation.py
```diff
@@ -249,9 +293,18 @@ def download_and_process(
     Raises
     ------
     ValueError
-        If any given workflow name is unknown.
+        If a mandatory argument is missing, if both ``workflows`` and ``workflow`` are
+        given, or if any given workflow name is unknown.
     """
-    workflows = [workflows] if isinstance(workflows, str) else list(workflows)
+    workflows = _resolve_workflows_argument(workflows, workflow)
+    for _name, _value in (
```
chrisjwin: Why this instead of just making them mandatory arguments of the function as they were before?

reskit/workflow_manager.py
```diff
@@ -850,6 +903,9 @@ def distribute_workflow(
     assert isinstance(placements, pd.DataFrame)
     assert ("lon" in placements.columns and "lat" in placements.columns) or ("geom" in placements.columns)
 
+    # work on a copy, the caller's placements table must not be changed by this function
+    placements = placements.copy()
```
chrisjwin, line 907: Should this copy not be creared right away a few lines up at `placements = self.placements`?

test/01_util/util/test_topography.py
```diff
@@ -1,12 +1,40 @@
...
+import pytest
 
 from reskit import TEST_DATA
 from reskit.util.topography import visibility_from_topography
 
 
 def test_visibility_from_topography():
     output = visibility_from_topography(lon=6.0, lat=51, elevation_raster=TEST_DATA["DEM-like.tif"])
-    assert np.isclose(output["visibility"].values.mean(), 0.2888095238095238)
+    assert np.isclose(output["visibility"].values.mean(), 0.27595238095238095)
```
chrisjwin: What is the reason for the value change? Maybe comment or so?

test/05_workflows/csp_workflows/test_csp_worklfow_manager.py
```diff
@@ -1157,7 +1157,7 @@ def pt_PTRWorkflowManager_calcCFs() -> PTRWorkflowManager:
     # wfm.placements['aperture_area_m2'] = 1E5
     # wfm.placements['land_area_m2'] = 1E5 / 0.3
 
-    # wfm._time_index_ = pd.date_range("2014-12-31 23:30:00", periods=100, freq="H")
+    # wfm._time_index_ = pd.date_range("2014-12-31 23:30:00", periods=100, freq="h")
```
chrisjwin, line 1160: What is happening here? Do we still need the commented out lines, if so we should uncomment them, else we could delete them? It feels a bit strange though to correct commented-out code? :)

chrisjwin (changes requested):
Thanks for cleaning up and increasing test coverage!! A few minor remarks

reskit/parameters/parameters.py
phil-fzj: ignore_index is not a parameter of pd.read_csv()

reskit/solar/core/system_design.py
phil-fzj: reverted

reskit/util/input_preparation.py
phil-fzj: reverted

reskit/util/input_preparation.py
phil-fzj: reverted

test/01_util/util/test_topography.py
phil-fzj: added

test/05_workflows/csp_workflows/test_csp_worklfow_manager.py
phil-fzj: kept as discussed

reskit/util/input_preparation.py
phil-fzj: reverted

phil-fzj:
@chrisjwin I adapted the code as discussed

phil-fzj:
But merge #326 first, as its a merge in the branch of this PR

chrisjwin (approved):
Great thanks a lot

PR #323, opened 2026-09-02
Stop the PV workflow from reading unused snow data

phil-fzj:
## Problem

`openfield_pv_era5` read four snow variables which no calculation uses, and
`download_and_process()` does not download them. Every simulation on ERA5 data prepared that
way stopped in `NCSource.load()` with a bare `AssertionError`.

The bundled test data contains the snow variables, so the suite never showed this.

## Change

A measurement decided the direction: with and without the four variables, every other output
variable is identical. The workflow needs no snow data, so the four names are removed from its
`read()` list.

Snow support itself stays usable. The four `sload_snow_*` loaders of `Era5Source`, their tests
and the test data are unchanged, and the download and tiling keep their snow entries, so a user
can still prepare snow data on purpose.

The bare `assert` in `NCSource.load()` now raises a `ResError` which names the absent variable
and lists the variables the source does have.

## For the reviewer

New tests compare, for every registered ERA5 workflow, the variables of its `read()` calls
against the variables its `depends_on` entry prepares, and check that every raw variable has a
tile label. Two more tests keep snow support usable and keep unused snow data out of the
workflow dependencies. None of them needs a download.

## Verification

338 passed, 5 skipped.


test/01_util/util/test_input_preparation.py
```diff
@@ -158,3 +163,120 @@ def test_ERA5_NC_TO_TILE_LABEL_ws100_label():
...
+    """
+    for cds_name in _SNOW_CDS_NAMES:
+        assert cds_name in era5_variables
+        assert cds_name in Era5Source.CDS_TO_NC_NAME
+        assert Era5Source.CDS_TO_NC_NAME[cds_name] in _ERA5_NC_TO_TILE_LABEL
+
+    assert set(Era5Source.raw_passthrough_variables(_SNOW_CDS_NAMES)) == {"asn", "rsn", "sd", "sf"}
+
+
+def test_no_workflow_depends_on_the_snow_variables():
```
chrisjwin: At least this will change... Maybe revise this whole PR and adapt it slightly with regards to the coming solar workflow with snow?

chrisjwin: Instead of focusing on snow so much, the differentiation should be rather workflow-specific between used and unused variables?

chrisjwin (changes requested):
Please review this as it will lead to issues with the solar workflow under preparation

test/01_util/util/test_input_preparation.py
phil-fzj: removed snow tests

chrisjwin (approved):

PR #324, opened 2026-09-02
Download ERA5 in the CF compliant format the CDS supports

phil-fzj:
## Problem

`era5_downloader()` asked for `data_format="netcdf_legacy"` and sent it under the obsolete
`format` request key. The CDS answers every such request with

```text
The 'netcdf_legacy' format is deprecated and no longer supported.
```

so no ERA5 download worked.

## Why the fix is more than a string

A live download of five variables shows what the CF compliant `netcdf` format does differently:

- it names the time axis `valid_time`, not `time`;
- it adds the coordinates `number` and `expver`;
- it returns a ZIP archive with one member per `stepType` group, even when the request asks for
  `download_format="unarchived"`.

The archive has a cause worth knowing. An instantaneous field such as `t2m` is valid at its time
stamp; an accumulated field such as `ssrd` is valid over the hour before it. A CF compliant file
must mark that difference on the time axis, and one time axis cannot carry both marks.
`netcdf_legacy` put every field on one axis and lost the difference. `download_format="unarchived"`
therefore cannot prevent the archive — it only suppresses an archive around a single file.

Three further live requests measure this: two instantaneous variables give one file, two
accumulated variables give one file, and one of each gives an archive with the members
`data_stream-oper_stepType-instant.nc` and `data_stream-oper_stepType-accum.nc`. The `stepType`
mix drives the split, not the variable count.

## Change

- Send `data_format` and `download_format` under the current request keys and default to
  `"netcdf"`. `"netcdf_legacy"` stays available for an archive which still holds that format.
- Normalise the answer in one place. `_normalize_era5_download()` merges a ZIP archive into one
  file and renames `valid_time` to `time`. A file which already uses `time` is not rewritten, so
  no download pays for a step it does not need.
- `_time_var_name()` lets `_nc_years()`, and therefore `era5_tiler()`, read both names.
  `Era5Source` passes `time_name=("time", "valid_time")`, and `NCSource` now accepts a sequence
  of time names and takes the first one the files contain. A single name behaves as before.
- Give `compat` explicitly to `xr.merge()`, because xarray is about to change that default.

## Verification

9 new tests, of which 7 fail without the fix. The other two are guards: a legacy answer must stay
untouched, and `preprocess_era5_data()` already handled `valid_time`. A fake CDS client keeps
every test offline.

Suite: 326 passed, 5 skipped.


chrisjwin:
@phil-fzj FYI check pipeline

phil-fzj:
@chrisjwin FYI pipeline is fixed

chrisjwin (approved):

PR #325, opened 2026-09-02
Repair the CSP area determination

phil-fzj:
Two defects in `PTRWorkflowManager.determine_area()`. They sit in the same function, so they
share one branch.

## 1. The discarded `drop()` results

The legacy `"area"` and `"area_m2"` columns were dropped, but the result of `DataFrame.drop()`
was discarded. The legacy column stayed in `self.placements` and reached the simulation output.

## 2. Two documented area inputs always crashed

The bodies of the last two branches were exchanged. Each branch read the area column which is
absent in that branch, so the documented inputs `aperture_area_m2` and `land_area_m2` always
raised a `KeyError`.

## Change

- Assign the result of both `drop()` calls.
- Exchange the two branch bodies.
- Make the `"area_m2"` block an `elif` of the `"area"` block, so placements which carry both
  columns give a deterministic result.
- Document the order of the area columns and the meaning of the solar field density.

## For the reviewer

Tests cover all four documented area inputs, both areas given together, the precedence of
`"area"` over `"area_m2"`, idempotence of a second call, and the absence of `"area_m2"` from the
CSP workflow output.

This branch touches the same lines as `maintenance/quick-wins-and-bugs`. Expect a conflict in
`csp_workflow_manager.py` if both are merged; the changes are independent and both are wanted.

## Verification

329 passed, 5 skipped.


reskit/csp/workflows/csp_workflow_manager.py
```diff
@@ -90,22 +94,22 @@ def determine_area(self):
         if "area" in columns and not "aperture_area_m2" in columns and not "land_area_m2" in columns:
```
chrisjwin: Could it be more meaningful to enforce a clear area parameter name and force the user to preprocess in case?

chrisjwin (changes requested):
Open to discuss if this is actually what is needed or if we do not mess up the package a little?

reskit/csp/workflows/csp_workflow_manager.py
chrisjwin: As a compromise, the other area types could also be captured with a deprecation warning and removed sometime soon?

phil-fzj: fixed. depreacated "area" and "area_m2".

chrisjwin (approved):
Great so far, small remaining points will be addressed when we review the csp module as a whole

chrisjwin (approved):

PR #326, opened 2026-09-02
Fail the suite on upstream deprecations that RESKit triggers

phil-fzj:
## Base

This branch targets `maintenance/quick-wins-and-bugs`, not `dev`. That branch contains the code
fixes which the gate below protects, so the gate can only be green on top of it. Merge that
branch first.

## Problem

Several upstream projects announced removals which RESKit code triggers. The fixes were made,
but nothing stopped the next one from appearing: the suite was green with 12 warnings of this
kind, and no CI job promoted them to errors.

## Change 1 — a selective warning gate

`filterwarnings` in `[tool.pytest.ini_options]` of `pyproject.toml`. Every CI job runs pytest and
therefore reads it, so no workflow file changes.

The gate names five warnings and turns only those into errors:

| Warning | Upstream change |
|---|---|
| `pvlibDeprecationWarning` | pvlib 0.13.1 renamed `singleaxis(apparent_azimuth=)` to `solar_azimuth` |
| xarray `drop` FutureWarning | `Dataset.drop()` is replaced by `Dataset.drop_vars()` |
| xarray `join` FutureWarning | the merge default changes from `join="outer"` to `join="exact"` |
| `logging.warn` DeprecationWarning | Python 3.13 removes the name |
| `PytestReturnNotNoneWarning` | pytest will make a returning test an error |

Everything else stays a warning on purpose. Third-party noise is not ours to repair, and RESKit
raises its own deprecations for its own users; a blanket `error` would fail the suite on both.

A filter only sees a line which a test executes, and the suite executes only one of the four
pvlib call sites. `test/08_dependencies/test_no_deprecated_apis.py` therefore reads the syntax
tree of every RESKit module and rejects the removed pvlib keyword. Add a guard there for each
further removal the suite cannot reach.

## Change 2 — twelve tests returned a value

pytest answers a test which returns a value with `PytestReturnNotNoneWarning`. The cause here is
a factory which carries a test name: five of the twelve build a workflow manager, assert its
state and return it, and other tests call them to get that manager. Seven more returned a manager
which no caller read.

Each of the five keeps its body under a `_make_` name, and a test of the old name calls it. The
assertions stay in the test report and the test count does not change. The seven dead returns are
removed.

## Verification

Each gate entry was checked by restoring the file from before its fix and running the tests which
touch it:

| Restored file | Result |
|---|---|
| `csp_workflow_manager.py` | pvlib and `logging.warn` errors — 3 failed, 2 errors |
| `csp/workflows/workflows.py` | `drop` error — 1 failed |
| `util/create_LRA.py` | `join` error — 7 failed |
| `test_wind_workflow_manager.py` | return-not-none error on both wind factories |

Suite: 387 passed, 3 skipped. Base branch on the same day: 386 passed, 3 skipped.

### DIFF
--- a/test/08_dependencies/test_no_deprecated_apis.py
+++ b/test/08_dependencies/test_no_deprecated_apis.py
@@ -4,12 +4,19 @@ The runtime gate in ``pyproject.toml`` turns such a deprecation warning into an
 error. That gate only sees a line which a test executes. The tests here read the
 RESKit source instead, so they also cover the lines which no test executes.
 
+A guard which removes a keyword makes RESKit give that argument by position. A
+position holds only while the upstream signature holds, so each such guard needs a
+second guard on the upstream signature.
+
 Add a guard here when an upstream project announces a removal which RESKit code
 touches. Record the upstream version in the docstring of the guard.
 """
 
 import ast
+import inspect
 from pathlib import Path
+
+import pvlib
 
 import reskit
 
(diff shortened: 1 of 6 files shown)


test/08_dependencies/test_no_deprecated_apis.py
```diff
@@ -0,0 +1,55 @@
...
+    return sorted(
+        node.lineno
+        for node in ast.walk(tree)
+        if isinstance(node, ast.Call)
+        for argument in node.keywords
+        if argument.arg == keyword
+    )
+
+
+def test_no_call_uses_the_pvlib_apparent_azimuth_keyword():
```
chrisjwin, line 47: This makes sense! Would it be possible also to do the inverted test that the pvlib function keeps either apparent_azimuth or solar_azimuth in this very same argument position? With positional arguments, we always run a certain risk of a shifted order which may lead to hard-to-track (or even unnoticed) errors as long as the datatypes and value ranges are somewhat comparable

chrisjwin (changes requested):
Important for consistency, maybe consider adding the inverse test as well to make sure potional order does not change?

phil-fzj:
@chrisjwin comments resolved

chrisjwin (approved):
Great!

PR #328, opened 2026-09-03
Remove the dead combine_mode option of create_LRA

phil-fzj:
## Base

This branch is the top of a stack. It targets `fix/deprecation-warnings`, which targets
`maintenance/quick-wins-and-bugs`, which targets `dev`. Merge the stack from the bottom.

A `dev` base is not possible: `maintenance/quick-wins-and-bugs` already changed the two
`xr.merge()` calls which this branch replaces, so a `dev` base conflicts on both lines.

## Problem

`create_LRA.py` declared `combine_mode` in six function signatures, threaded it through the whole
call chain, documented it, and offered it on the command line as `--combine-mode` with the choices
`auto`, `merge` and `combine_by_coords`. No line ever read the value. `load_era5_year()` always
called `xr.merge()`, and the tile combination at the end of `compute_dni_year()` did the same. The
docstring of `load_era5_year()` names the intended rule, but the dispatch was never written.

## Change

The option is removed. Both tile combinations now call one new helper, `_combine_tiles()`, which
combines the tiles by their coordinates.

The option has no purpose. Measured on xarray 2026.4.0, `xr.merge(compat="no_conflicts",
join="outer")` and `xr.combine_by_coords` return the same result for every tile layout this module
can meet:

| Tile set | `xr.merge` | `xr.combine_by_coords` |
|---|---|---|
| Same variable, disjoint coordinates | same result | same result |
| Different variables, same or disjoint coordinates | same result | same result |
| Ragged tile set: a hole, or tiles of unequal shape | works | works |
| Same variable, overlapping coordinates | unions the overlap in silence | raises `ValueError` |

One case separates the two functions, and there the stricter one is correct. Tiles which overlap
on the same variable are an input error, and a silent union hides it. So one code path covers
every input, and a choice between two functions gives the user nothing to decide.

`_combine_tiles()` also converts a `DataArray` to a `Dataset` before it combines.
`compute_dni_year()` collects `DataArray` objects, and `xr.combine_by_coords` returns a
`DataArray` for `DataArray` input, which would change the return type of that function.

## For the reviewer

`create_LRA.py` is new and unreleased, so the removed parameter keeps no alias.

The change deletes an option, so it adds no test. The test of the command-line defaults loses its
assertion on the removed option.

### DIFF
--- a/reskit/util/create_LRA.py
+++ b/reskit/util/create_LRA.py
@@ -33,7 +33,7 @@ import argparse
 import glob
 import logging
 from pathlib import Path
-from typing import Iterable, Literal, Optional
+from typing import Iterable, Optional
 import xarray as xr
 from tqdm.auto import tqdm
 import numpy as np
@@ -41,9 +41,6 @@ import pandas as pd
 import geokit as gk
 
 LOG = logging.getLogger(__name__)
-
-
-CombineMode = Literal["auto", "merge", "combine_by_coords"]
 
 
 def _list_tiled_nc_files(base_path: Path, year: int, variable: str, zoom_level: int) -> list[Path]:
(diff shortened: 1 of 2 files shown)


chrisjwin (approved):
Great thanks for the fixes

Issue #226, opened 2026-01-26
Apply PEP8 naming standards


julian-belina:
### Purpose of the improvement

Currently, the naming of variables, modules and objects does not adhere to the PEP8 standard.

### Proposal

Adhere to the PEP8 standard to help new users find their way around more easily.

phil-fzj:
For reference, @chrisjwin and I agreed to the following:
- adhere to PEP8: `A style guide is about consistency. Consistency with this style guide is important. Consistency within a project is more important. Consistency within one module or function is the most important.`
- camel case for classes, snake case for functions and variable names, upper case for constants (guideline for new code additions).
- Some of our code differs in that regard. We can directly change everything that does not break the API.
- The goal is to maximize backward compatibility. However, we are open to break backward compatibility but only with sufficient transition periods with deprecation warnings

phil-fzj:
For RESKit specifically, changes would mostly apply to DAC, CSP and geothermal workflows.
Here, we decide to break backward compatibility but only with a transition period. 
If workflow names need to be changed we will have 2 separate functions. One with the old naming and one with the updated. Using the old one will trigger a deprecation warning for a sufficiently long transition period. The new one is a wrapper only but will pass a no deprecation flag.
Deeper level functions can be changed immediately.
This change should trigger a v0.6.0 release.

Deprecation period can end with a release of v1.0.0 in a few months

FYI @julian-belina

phil-fzj:
done with #330, #331, #332

PR #330, opened 2026-09-08
Apply PEP 8 naming to classes, functions and methods (#226)

phil-fzj:
Renames the identifiers in reskit/ that break the agreed convention: CamelCase for classes, snake_case for functions and methods. Acronyms are lower case inside snake_case names, which is the style already used by the wind, solar and DAC workflows and by the ERA5 preparation code.

**Scope of this PR is limited to classes, functions and methods.** Variable and constant names keep their current spelling and follow later, per module.

Names that are not part of the public API are renamed outright, together with their call sites:

  - classes: EGS_workflowmanager -> EGSWorkflowManager, dataset_handler -> DatasetHandler, gringarten -> Gringarten
  - the 12 metric functions in csp/validation/validate.py
  - the 8 internal cost functions in wind/economic/offshore_cost_model.py
  - waterDepthFromLocation -> water_depth_from_location, distanceToCoastline -> distance_to_coastline, get_tile_XY -> get_tile_xy, _calculate_DNI -> _calculate_dni
  - the 6 Gringarten and 2 DatasetHandler methods

Names that users call keep working. Each old name stays available as a wrapper that raises a DeprecationWarning and delegates to the new name. The wrappers are removed in v1.0.0:

  - CSP_PTR_ERA5 -> csp_ptr_era5
  - CSP_PTR_ERA5_specific_dataset -> csp_ptr_era5_specific_dataset
  - EGSworkflow -> egs_workflow
  - TurbineLibrary -> turbine_library
  - calculateSpecificOffshoreCapex -> calculate_specific_offshore_capex

A wrapper cannot cover a workflow name that is passed as a string, so util/input_preparation.py keeps the old keys of depends_on resolvable and warns when one of them is used.

The example notebooks in docs/examples/ and examples/ now use the new names. test/01_util/util/test_deprecated_names.py checks that every old name still works, warns, and returns what the new name returns.

Method renames on PTRWorkflowManager, EGSWorkflowManager, NCSource and SolarWorkflowManager are left for v1.0.0, so that user code breaks once rather than twice.


phil-fzj:
@chrisjwin @julian-belina this would be my proposal. Feel free to add suggestions or feedback.

julian-belina:
@chrisjwin, I checked this PR and added two more on top. 

The first, https://github.com/FZJ-IEK3-VSA/RESKit/pull/331, fixes the original problem that prompted my motivation for a renaming solution. A simple reordering of the imports according to the isort linter in Ruff caused the entire test suite to crash. This has been resolved by a more explicit import method in https://github.com/FZJ-IEK3-VSA/RESKit/pull/331.

This revealed the original issue that some modules and classes had the same name, enabling the import to be broken simply by reordering them. This has been fixed in PR https://github.com/FZJ-IEK3-VSA/RESKit/pull/332 by applying proper PEP8 names to all modules. However, to maintain backwards compatibility, we use sys.module.

https://github.com/FZJ-IEK3-VSA/RESKit/pull/332/changes#diff-aba458c0925557b1d7ed4b64d677a1eca5d89d2f1eac38eed8bf93d63b33a66aR131.

This dictionary relinks the old module paths to the new ones (https://github.com/FZJ-IEK3-VSA/RESKit/pull/332/changes#diff-aba458c0925557b1d7ed4b64d677a1eca5d89d2f1eac38eed8bf93d63b33a66aR33). This enables backwards compatibility while using proper names. The naming confusion is also completely resolved with the 1.0 release.

Regarding the internal variable names, we currently don't have a standard. My proposal would be to enforce pep8 for new code and allow the step-by-step replacement of old variables when the code is reworked or extended. However, I would be open to debate on this point.

julian-belina (approved):

PR #340, opened 2026-09-16
Speed up the Gaussian power curve convolution by ~4-10x

phil-fzj:
`PowerCurve.convolute_by_gaussian` was the largest single cost in the wind workflows after reading the input data.

The cost is paid per distinct power curve on every workflow call, and is independent of the number of turbines and of the weather source.

Two changes, both numerically neutral (max abs difference ~1e-15):

- Evaluate the Gaussian directly instead of calling `scipy.stats.norm.pdf`, which re-validates its arguments and goes through the `rv_continuous` machinery on each of the `_steps` calls.

- Only evaluate it within +-8 standard deviations of its centre. The kernel is narrow (a std of 0.4 m/s at the top of the range with the workflows' scaling=0.01, base=0.0) against a 40 m/s sweep, so nearly all of the 16 million evaluations were of numbers indistinguishable from zero. This turns the algorithm from O(`_steps`^2) into O(`_steps` x band).

Measured on a synthetic 300 W/m2 curve at the default `_steps=4000`: 210 ms -> 20 ms for the workflows' (0.01, 0.0), 208 ms -> 36 ms for the default (0.06, 0.1).


phil-fzj:
Total workflow runtime (median of 3 runs):

| placements | source  | before | after  | saving | speedup |
|-----------:|---------|-------:|-------:|-------:|--------:|
| 20         | netCDF4 | 3.79 s | 2.47 s | 1.32 s | 1.5x    |
| 1000       | netCDF4 | 10.93 s | 8.90 s | 2.03 s | 1.23x   |

Note that speedup scales with the number of distinct power curves.

Also note that at 1000 placements the convolution is no longer the dominant step.

chrisjwin (approved):
Thanks, great! I have taken a note in the new discussions tab on our further idea of storing preconvoluted power curves in a dict here https://github.com/FZJ-IEK3-VSA/RESKit/discussions/346

PR #341, opened 2026-09-16
Load the wind cost models on first use, not on import

phil-fzj:
`import reskit` took 2.90 s and 456 MB, of which 1.85 s and 390 MB was WISDEM. It arrived through `reskit.wind.economic.onshore_cost_model`, which `reskit/wind/__init__.py` imported eagerly, and the offshore model pulls the same module in for `onshore_tcc`. Nothing in the simulation workflows calls either of them. Inside WISDEM, 1.3 s of the load is `jsonschema` and `rfc3987_syntax`, used by its input validation.

That is per process. 
The three exported names are resolved on first attribute access instead (PEP 562), and bound into the module's globals once resolved. Accessing any of them behaves as before and loads WISDEM at that point; importing the submodules directly is unaffected.

    import reskit:  2.90 s -> 0.81 s,  456 MB -> 311 MB


chrisjwin (approved):
Very good idea! I have taken a note on potential further improvements by extracting the core algorithms into RK directly in our discussions section, this would avoid the loading time also for script including wind economics, but more importantly reduce dependencies

PR #342, opened 2026-09-16
vectorise read and ws correction

phil-fzj:
Stacked on #340. The base branch is `perf/power-curve-convolution`, and all numbers
below are measured against it. Please review #340 first.

After the convolution fix in #340, the two biggest remaining costs in the wind
workflow were in code shared by both weather sources. This PR removes them. The
capacity factors are bit-identical before and after on both sources.

## What changed

1. `NCSource.get` fitted a new `RectBivariateSpline` for every time step, which is
   35 040 spline fits per workflow call over a year of hourly data with four
   variables. For the bilinear case the spline is just bilinear interpolation, and
   the weights depend only on the grid and the placements, not on the values. They
   are now computed once and applied to all time steps together. The cubic case
   still uses the loop, because a cubic spline is not local.

2. The `ws_bins` wind speed correction made one pass over the whole
   (time x location) array per bin, and the table used by the 2025 workflow has 320
   bins. The bins are sorted and do not overlap, so a single `searchsorted` now
   assigns every wind speed to its bin. If bins do overlap the result depends on the
   order they are applied in, so that case is detected and still uses the loop.

3. Fixes a bug in (1) and adds tests. The new interpolation extrapolated for points
   outside the extraction window. `RectBivariateSpline` does not do this, it clamps
   them to the edge of the window, and `NCSource.get` relies on that for locations
   above the highest latitude in the grid. The two differed by up to 97 W/m2 for
   such points. Wind workflow placements are always well inside the window, so no
   workflow result was affected.

## Performance

`wind_era5_PenaSanchezDunkelWinklerEtAl2025` over one ERA5 tile, one year of hourly
data, warm page cache, median of three runs after a warmup.

| | | #340 | this PR | speedup |
| --------- | ------- | ------ | ------ | ----- |
| workflow, 20 placements | netCDF4 | 2.47 s | 0.36 s | 6.9x |
| workflow, 20 placements | Zarr | 0.90 s | 0.33 s | 2.7x |
| workflow, 1000 placements | netCDF4 | 8.90 s | 2.72 s | 3.3x |
| workflow, 1000 placements | Zarr | 7.38 s | 2.69 s | 2.7x |
| `WorkflowManager.read` only, 1000 placements | netCDF4 | 4.88 s | 0.49 s | 10.0x |
| `WorkflowManager.read` only, 1000 placements | Zarr | 3.38 s | 0.47 s | 7.2x |

Small runs improve as much as large ones because the spline fitting happened once
per time step, regardless of how many placements were simulated.

The two functions on their own, on arrays of the size the workflow gives them:

| | before | after | speedup | difference |
| --- | ------ | ----- | --- | --------- |
| bilinear, 8760 time steps, 1000 points, one variable | 358 ms | 59 ms | 6.1x | 1.07e-14 max abs |
| `ws_bins`, 8760 x 1000 array, 320 bins | 2295 ms | 295 ms | 7.8x | none, bit-identical |

Peak memory is unchanged: 1725 -> 1729 MB on netCDF4 and 2318 -> 2302 MB on Zarr at
1000 placements.

## Testing

- Ran the full workflow on both sources before and after, 200 placements.
  `np.array_equal` on the capacity factors returns `True`. Repeated after the fix in
  (3).
- `pytest test/` without the notebook examples: 465 passed, 3 skipped. `ruff check`
  and `ruff format --check` pass.
- Added three tests: the new interpolation against a per-time-step
  `RectBivariateSpline`, including points outside the window, which is what found
  the bug in (3); the `ws_bins` lookup against the loop it replaces; and the
  overlapping-bin case, which had no test before. Also added one assertion for
  `interpolation="cubic"`, which was untested.


phil-fzj:
Against a testrun with 1000 placements this results in a speedup of ~4x for the complete workflow runtime

| phase    | cache | source | dev    | stacked PR| speedup | bytes read |
|----------|-------|--------|-------:|-----------------:|--------:|-----------:|
| workflow | cold  | netcdf | 69.11s | 16.49s           | 4.19x   | 162M       |
| workflow | cold  | zarr   | 62.66s | 14.47s           | 4.33x   | 154M       |

FYI: @chrisjwin

reskit/wind/core/windspeed_correction.py
```diff
@@ -57,13 +57,35 @@ def correction_function(x):
...
+        _disjoint = bool(np.all(_rights[:-1] <= _lefts[1:]))
+
         def correction_function(x):
             # x is numpy array. modify x based on ws_bins
-            corrected_x = x.copy()
-            for ws_bin, factor in ws_bins_correction.items():
-                mask = (x >= ws_bin.left) & (x < ws_bin.right)
-                corrected_x[mask] = x[mask] * (1 - factor)
-            return corrected_x
+            if not _disjoint:
```
chrisjwin: replace by check that not disjoint else raise?

chrisjwin (changes requested):
Minor update

chrisjwin (approved):
Great

PR #344, opened 2026-09-16
Fix bounds covering afull data extent returning a 6x6 corner

phil-fzj:
Opening a weather source with bounds that cover the whole dataset returned a 6x6 cell window from one corner instead of the full grid — no error, no warning, the simulation just ran on the wrong data. Both Era5Source and Era5ZarrSource are affected; the code is in their shared parent, NCSource.

The window was found by locating the first cell outside the bounds, with argmin over the inside-mask. When every cell is inside, the mask is all True, argmin has no match and falls back to 0, and the window collapses to the index padding.

This takes the first and last cell inside the bounds directly instead. Two smaller things come with it: bounds that miss the data entirely now raise instead of silently clipping, and the start index can no longer go negative.

One test expectation changed. test_MerraSource___init__ had the old behaviour baked in — the Aachen extent it builds spans 5.47 to 6.92, which contains all three longitudes of the test file, but it asserted only two came back. It now expects three.

To check it on a real tile:

for bounds in [(-1.0, 40.0, 23.5, 56.75),   # inset by 1 degree
               (-2.0, 39.0, 24.5, 57.75),   # exactly the data extent
               (-5.0, 35.0, 30.0, 60.0)]:   # larger than the data extent
    src = Era5Source(TILE_PATH, bounds=gk.Extent(*bounds).castTo(4326))
    src.sload_surface_pressure()
    print(bounds, "->", src.data["surface_pressure"].shape)
All three give (8760, 76, 107) on a 107x76 tile; the last two gave (8760, 6, 6) before. Regression tests for all three cases are in test_Era5Source.py. Full suite passes.


chrisjwin (approved):
Great!

PR #349, opened 2026-09-17
346 location specific args in execute workflow iteratively

### LINKED ISSUE #348: allow location-specific workflow args also when using execute_workflow_iteratively

The current execute_workflow_iteratively() function cannot deal with location specific workflow argument values as long as they are NOT columns of the placements df. 

### Proposal

Workflow specific args need to be selected/masked analog to the placements row selection

chrisjwin:
Allow location specific args in execute_workflows_iteratively(). Preserves backward compatibility via additional arg for location specific parameters


reskit/workflow_manager.py
```diff
@@ -1020,28 +1035,88 @@ def execute_workflow_iteratively(
...
-    if not weather_path_varname in placements.columns:
+    if weather_path_varname not in placements.columns:
         weather_path = workflow_args[weather_path_varname]
         placements = get_dataframe_with_weather_tilepaths(placements=placements, weather_path=weather_path, zoom=zoom)
-    if not "RESKit_sim_order" in placements.columns:
-        placements["RESKit_sim_order"] = range(len(placements))
-
-    # remove output saving for the iterative function execution of sub dfs
-    workflow_args.update({"output_netcdf_path": None})
+        tilepaths = np.asarray(placements["source"])
```
phil-fzj: You added a new case where the weather path is already a column in the placements table (line 1038). But the progress message a few lines further down (line 1116) still reads placements['source'], which only exists in the other case. I tried it: it crashes with KeyError: 'source' on the first tile.

reskit/workflow_manager.py
```diff
@@ -1020,28 +1035,88 @@ def execute_workflow_iteratively(
...
+                # do not save anything in this sub iteration
+                _workflow_args[_arg] = None
+            elif _arg == weather_path_varname:
+                # set the current weather path tile path - NOTE that this arg can be either here or in location-specific args, hence again below
+                _workflow_args[_arg] = tilepath
+            else:
+                # we can use it as it is, it is a standard "global" arg across all locs
+                pass
+        # now iterate over the location-specific args and select only those values that apply to the tile subset of the placements df
+        for _arg, _val in location_specific_workflow_args.items():
```
phil-fzj, line 1147: The block that handles the weather path inside location_specific_workflow_args (lines 1077-1081) can never run. The check above it requires the weather path to be in the normal arguments, and another check forbids it being in both places. So either drop that block or loosen the check.

phil-fzj (changes requested):
also, please add tests

reskit/workflow_manager.py
chrisjwin: thank you, I aligned the placements output now independent of the source of the weather data path var

reskit/workflow_manager.py
chrisjwin: thank you. I widened above check to make sure it is in ONE of the two, then later assert that it is not duplicated

phil-fzj (approved):

